PR #3, opened 2019-12-17
Create inline security groups and conditionals to manage them

tthyer:
This creates two security groups. Which one gets used depends on whether the end user choose public or private subnet. I am making my best guess about what is needed by the majority of cases. Please review whether PublicSubnetSecurityGroup is too permissive.


tthyer:
@zaro0508 "Are there plans to support the other common use case of Public and open other ports?" Yes, but I would like to discuss with you today the best way to go about that. I'm tempted to make these separate templates (private/public), then use the CFN macros for the open ports.

PR #6, opened 2019-12-19
Hard code list of available instance types for linux and windows

tthyer:
instance type list hard-coded, not using the include


ec2/sc-ec2-linux-ra.yaml
```diff
@@ -32,16 +32,158 @@ Parameters:
...
+      - r5n.8xlarge
+      - r5n.12xlarge
+      - r5n.16xlarge
+      - r5n.24xlarge
       - t2.nano
       - t2.micro
       - t2.small
       - t2.medium
       - t2.large
+      - t2.xlarge
```
zaro0508: remove these t2 as well?

PR #7, opened 2019-12-19
Linux distro

tthyer:
Allows a choice of linux distribution to user.

### DIFF
--- a/ec2/sc-ec2-linux-ra.yaml
+++ b/ec2/sc-ec2-linux-ra.yaml
@@ -22,6 +22,12 @@ Metadata:
         default: EC2 Instance Type
       LinuxDistribution:
         default: Linux Distribution
+Mappings:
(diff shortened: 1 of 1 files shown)


ec2/sc-ec2-linux-ra.yaml
```diff
@@ -12,16 +12,22 @@ Metadata:
...
       EC2InstanceType:
         default: EC2 Instance Type
-      LatestAmiId:
-        default: SSM key to the latest Amazon linux AMI
+      LinuxDistribution:
+        default: Linux Distribution
+Mappings:
+  LinuxDistributions:
+    Ubuntu:
+      AMIID: ami-04b9e92b5572fa0d1
```
zaro0508: If these are AWS AMIs they are not guaranteed to be around.  We should save a copy of these in our account.

PR #13, opened 2019-12-21
Fix ImportValue in EC2 product templates

tthyer:
Update ImportValue in ec2 products to use ScipoolInternalVPC


tthyer:
@zaro0508 Changed all the vpc import references to the internalpoolvpc exports in bc2f99d.

PR #15, opened 2020-01-02
WIP explode

tthyer:
This is a WIP pr. The idea is to allow a list of ports to open on an EC2 instance.


zaro0508:
looks good to me

PR #17, opened 2020-01-03
create a S3 synapse bucket service catalog item

zaro0508:


s3/sc-product-s3-synapse.yaml
```diff
@@ -0,0 +1,48 @@
...
+        - Description: baseline version
+          Info:
+            LoadTemplateFromURL: !Sub '${RepoRootURL}s3/sc-s3-synapse-ra.yaml'
+          Name: !Ref 'ProductVersionName'
+  Associates3:
+    Type: AWS::ServiceCatalog::PortfolioProductAssociation
+    Properties:
+      PortfolioId: !Ref 'PortfolioId'
+      ProductId: !Ref 'scs3product'
+  constraintec2linux:
```
tthyer: should have more appropriate name

zaro0508: this was just copied from https://github.com/Sage-Bionetworks/scipoolprod-sc-lib-infra/blob/master/s3/sc-product-s3-private.yaml#L39

tthyer: understood but looks like a mistake on their part

PR #25, opened 2020-01-09
Add Deny on unencrypted objects back to BucketPolicy in encrypted bucket

tthyer:
This adds back in the explicit Deny on trying to upload objects that are not encrypted. @zaro0508, please see if this whole policy makes sense to you.


tthyer:
@zaro0508 just debugging, but wanted to verify that my alteration/expansion makes sense to you

PR #26, opened 2020-01-14
Add Jumpcloud setup to ec2 template intended for use by ServiceCatalog

tthyer:
This adds init metadata to a ec2 instance template to set register the instance with Jumpcloud. 
- Parameter KeyPair is removed
- Parameter OwnerEmail is added
- No parameters were added for Jumpcloud. The SecureString parameters are pulled from SSM using awscli.

I have left several commented lines where I attempted to use Cloudformation Dynamic References. Unfortunately dynamic references cannot be used in the Parameters section nor in Init sections.

I was not able to fully test this template manually because even when assuming the Admin role I receive a message that I do not have permission to decrypt the strings.

### DIFF
--- a/ec2/sc-ec2-linux-jumpcloud.yaml
+++ b/ec2/sc-ec2-linux-jumpcloud.yaml
@@ -1,6 +1,11 @@
 AWSTemplateFormatVersion: '2010-09-09'
 Description: 'Service Catalog: EC2 Reference Architecture(fdp-1oc5f3ula).'
 Metadata:
+  cfn-lint:
+    config:
+      ignore_checks:
+        - E1022
+        - E3012
   AWS::CloudFormation::Interface:
     ParameterGroups:
       - Label:
@@ -307,7 +312,7 @@ Resources:
               owner: root
               group: root
             /lib/systemd/system/cfn-hup.service:
-              content: !Sub |
+              content: |
                 [Unit]
                 Description=cfn-hup daemon
 
@@ -342,7 +347,7 @@ Resources:
               command: !Join
                 - ''
                 - - "/usr/bin/curl --silent --show-error --header 'x-connect-key: "
-                  - "$(aws ssm get-parameters --names '/infra/JcConnectKey --with-decryption' | jq '.[] | .[0].Value')"
+                  - Fn::Transform: {"Name": "SsmParam", "Parameters": {"Type": "SecureString", "Name": "/infra/JcConnectKey"}}
                   - "' https://kickstart.jumpcloud.com/Kickstart | sudo bash"
         config_jc:
           commands:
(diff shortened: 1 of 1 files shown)


ec2/sc-ec2-linux-jumpcloud.yaml
```diff
@@ -0,0 +1,517 @@
...
+            02_associate_jc_systems:
+              command: !Join
+                - ''
+                - - "JC_SYSTEM_ID=$(sudo cat /opt/jc/jcagent.conf|jq -r '.systemKey'); "
+                  - "JC_USER_ID=$(/usr/bin/curl --silent --show-error -X GET https://console.jumpcloud.com/api/systemusers "
+                  - "-H 'Accept: application/json' -H 'Content-Type: application/json' -H 'x-api-key: "
+                  #- '{{resolve:ssm-secure:/infra/JcServiceApiKey:1}}'
+                  - "$(aws ssm get-parameters --names '/infra/JcServiceApiKey --with-decryption' | jq '.[] | .[0].Value')"
+                  - "' | jq -r '.results | .[] | select(.email==\""
+                  - !Ref OwnerEmail
```
zaro0508: I'm wondering if there's any way we can automatically get this from the current user SC user that's doing the provisioning?

ec2/sc-ec2-linux-jumpcloud.yaml
```diff
@@ -0,0 +1,517 @@
...
+          ToPort: '-1'
+          CidrIp: '0.0.0.0/0'
+      SecurityGroupEgress:
+        - Description: allow all outgoing
+          IpProtocol: '-1'
+          CidrIp: '0.0.0.0/0'
+  LinuxInstance:
+    Type: AWS::EC2::Instance
+    Metadata:
+      'AWS::CloudFormation::Init':
```
zaro0508: I think you need a `UserData` section because that's what calls this `Init`, something like this: https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/managed-ec2-linux-v1.j2#L274

ec2/sc-ec2-linux-jumpcloud.yaml
tthyer: @zaro0508 Not through CloudFormation, unfortunately. I looked into that when we were setting up the trial in IT Sandbox. However, maybe another awcli call here. aws sts has "get-caller-identity" which "Returns details about the IAM user or role whose credentials are used to call the operation." No idea what that's going to return when SC runs the template, though.

ec2/sc-ec2-linux-jumpcloud.yaml
tthyer: Thanks! I knew I forgot something.

ec2/sc-ec2-linux-jumpcloud.yaml
```diff
@@ -0,0 +1,517 @@
...
+                jq: []
+            -
+              yum:
+                jq: []
+          commands:
+            01_jumpcloud_agent:
+              command: !Join
+                - ''
+                - - "/usr/bin/curl --silent --show-error --header 'x-connect-key: "
+                  - "$(aws ssm get-parameters --names '/infra/JcConnectKey --with-decryption' | jq '.[] | .[0].Value')"
```
zaro0508: i'm guessing that this is not going to work because you need to setup the instance with an instance profile that has access to the SSM and the '/infra/JcConnectKey' Kms key.

ec2/sc-ec2-linux-jumpcloud.yaml
```diff
@@ -0,0 +1,517 @@
...
+                  #- '{{resolve:ssm-secure:/infra/JcConnectKey:1}}'
```
zaro0508: this won't work, directly resolving from the SSM in CFN usage is very limited.  look at "Resources that Support Dynamic Parameter Patterns for Secure Strings" in https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/dynamic-references.html

ec2/sc-ec2-linux-jumpcloud.yaml
tthyer: @zaro0508 I haven't found a way to do this. I can call `aws sts get-caller-identity`, but if this is done when a role has been assumed, as will be the case within ServiceCatalog, then only information about the assumed role is returned. I haven't found a way to reverse engineer this to get the original user.

ec2/sc-ec2-linux-jumpcloud.yaml
tthyer: Added in commit cb37c3f

ec2/sc-ec2-linux-jumpcloud.yaml
tthyer: Right. Anyway these commented out lines have been removed.

ec2/sc-ec2-linux-jumpcloud.yaml
```diff
@@ -1,6 +1,11 @@
 AWSTemplateFormatVersion: '2010-09-09'
 Description: 'Service Catalog: EC2 Reference Architecture(fdp-1oc5f3ula).'
 Metadata:
```
zaro0508: would you mind moving these ignore checks to the specific resource we want to ignore?

tthyer: @zaro0508 Did this, it's a bit repetitive.

PR #27, opened 2020-01-14
Add managed policy to hide template to protect secrets

tthyer:
This will prevent template from being viewable when a stack is examined.


tthyer:
@zaro0508 The policy that gives this access is AWSServiceCatalogEndUserFullAccess. I don't know which of the permissions in the policy are necessary for the ServiceCatalog to work correctly. While I agree with you it might be easier to just restrict all views of CloudFormation, I don't think it's trivial to work out what those are.

I think an alternate solution might be to have only an EndUser role rather than an EndUser group too. We don't want to give users these permissions all the time, just when using ServiceCatalog. Do you think that might work?

zaro0508:
I'm not sure how an EndUser role would help.  could you please elaborate?  I don't think we want to require SC users switch to a role just to use the SC.

tthyer:
@zaro0508 I was thinking give them the ability to assume the role for ServiceCatalog and not for other purposes.

PR #37, opened 2020-01-23
Create windows ec2 instance with jumpcloud integration

tthyer:
The primary purpose of this PR is creating a new Windows instance that includes cfn and jumpcloud scripts in the Init section and in UserData. That part is based on lessons learned from doing a similar task for the Linux Jumpcloud integration, as well as existing implementation in [aws-infra](https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/managed-ec2-win-v4.j2). The secondary purpose is to change the patch baseline to apply only security updates.

### DIFF
--- a/ec2/sc-ec2-windows-jumpcloud.yaml
+++ b/ec2/sc-ec2-windows-jumpcloud.yaml
@@ -205,7 +205,7 @@ Resources:
               PatchFilters:
                 - Key: PRODUCT
                   Values:
-                    - WindowsServer2016
+                    - WindowsServer2019
                 - Key: CLASSIFICATION
                   Values:
                     - SecurityUpdates
--- a/ec2/sc-ec2-windows-jumpcloud.yaml
+++ b/ec2/sc-ec2-windows-jumpcloud.yaml
@@ -1,0 +1,323 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Description: 'Service Catalog: EC2 Reference Architecture(fdp-1oc5f3ula).'
+Metadata:
+  AWS::CloudFormation::Interface:
+    ParameterGroups:
+      - Label:
+          default: General Configuration
+        Parameters:
+          - KeyPair
+      - Label:
+          default: Windows Instance Configuration
+        Parameters:
+          - WindowsInstanceType
+          - LatestAmiId
+    ParameterLabels:
+      KeyPair:
+        default: Key Pair
+      WindowsInstanceType:
+        default: Windows Instance Type
+      LatestAmiId:
+        default: SSM path for latest Windows AMI ImageId
+Parameters:
+  WindowsInstanceType:
+    AllowedValues:
+      - t3.nano
+      - t3.micro
+      - t3.small
+      - t3.medium
+      - t3.large
+      - t3.xlarge
+    Default: t3.micro
+    Description: Amazon EC2 Windows Instance Type
+    Type: String
+  KeyPair:
+    Description: Name of existing EC2 key pair for Windows Instances
+    Type: AWS::EC2::KeyPair::KeyName
+  LatestAmiId:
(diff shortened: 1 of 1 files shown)


tthyer:
@zaro0508 @ahayden I made the alterations to patching that I think we want, but I'm really not sure. I think what I've done means that all security updates, and security updates only, will be applied regardless of severity. But then how does ComplianceLevel affect that? Is the target product really WindowsServer2016? Please advise.

ec2/sc-ec2-windows-jumpcloud.yaml
```diff
@@ -0,0 +1,323 @@
...
+      - t3.medium
+      - t3.large
+      - t3.xlarge
+    Default: t3.micro
+    Description: Amazon EC2 Windows Instance Type
+    Type: String
+  KeyPair:
+    Description: Name of existing EC2 key pair for Windows Instances
+    Type: AWS::EC2::KeyPair::KeyName
+  LatestAmiId:
```
zaro0508: might want to change this to just `AmiId` because it's not the latest AMI.

ec2/sc-ec2-windows-jumpcloud.yaml
```diff
@@ -0,0 +1,323 @@
...
+      OperatingSystem: WINDOWS
+      ApprovalRules:
+        PatchRules:
+          - ApproveAfterDays: 0
+            ComplianceLevel: CRITICAL
+            PatchFilterGroup:
+              PatchFilters:
+                - Key: PRODUCT
+                  Values:
+                    - WindowsServer2016
```
zaro0508: The deault `LatestAmiId` refers to a windows 2019 server so i'm thinking we should make this match?  You'll probably need to make sure that WindowsServer2019 is a valid product first though.  https://docs.aws.amazon.com/systems-manager/latest/APIReference/API_DescribePatchProperties.html

ec2/sc-ec2-windows-jumpcloud.yaml
```diff
@@ -0,0 +1,323 @@
...
+        - Key: Description
+          Value: Service-Catalog-EC2-Reference-Architecture
+  WindowsPatchBaseline:
+    Type: AWS::SSM::PatchBaseline
+    Properties:
+      OperatingSystem: WINDOWS
+      ApprovalRules:
+        PatchRules:
+          - ApproveAfterDays: 0
+            ComplianceLevel: CRITICAL
```
zaro0508: add `EnableNonSecurity: false` like in sc-ec2-linux-jumpcloud.yaml?

ec2/sc-ec2-windows-jumpcloud.yaml
```diff
@@ -0,0 +1,323 @@
...
+                - ssm.amazonaws.com
+            Action:
+              - sts:AssumeRole
+  MaintenanceWindow:
+    Type: AWS::SSM::MaintenanceWindow
+    Properties:
+      Description: Maintenance window to allow for patching windows instances
+      AllowUnassociatedTargets: false
+      Cutoff: 2
+      Schedule: cron(* 17 * * ? *)
```
ahayden: I think this schedule is a good compromise between updates and availability... We need the machines to be up, but we also don't want to force reboots on people. An alternative would be to define `ScheduleTimezone` explicitly and set `Schedule` to the middle of the night, but its not clear to that users will park or not. Anyway, I defer to you if you think this gets the best coverage for the user base.

ec2/sc-ec2-windows-jumpcloud.yaml
tthyer: Changed in 143401c

ec2/sc-ec2-windows-jumpcloud.yaml
tthyer: Changed in 143401c

PR #44, opened 2020-01-29
Add OwnerEmail and change some text in sc-ec2-windows-jumpcloud

tthyer:
Minor windows fixes

### DIFF
--- a/ec2/sc-ec2-windows-jumpcloud.yaml
+++ b/ec2/sc-ec2-windows-jumpcloud.yaml
@@ -7,6 +7,7 @@ Metadata:
           default: General Configuration
         Parameters:
           - KeyPair
+          - OwnerEmail
       - Label:
           default: Windows Instance Configuration
         Parameters:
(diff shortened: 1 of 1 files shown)


ec2/sc-ec2-windows-jumpcloud.yaml
```diff
@@ -32,20 +35,25 @@ Parameters:
...
-    Type: AWS::SSM::Parameter::Value<AWS::EC2::Image::Id>
+    Type: AWS::EC2::Image::Id
     Default: ami-074545a6fb2313e2c
   VolumeSize:
     Description: The EC2 volume size (in GB)
     Type: Number
     Default: 30
     MinValue: 10
     MaxValue: 2000
+  OwnerEmail:
```
zaro0508: isn't this already a required tag in the SC?  would SC end user need to set this twice?

tthyer: That's right...but it's needed for 01_associate_jc_systems. Also doing it that way (requiring twice) in the linux template. I could rewrite these to pull OwnerEmail from tags as with TagRootVolume though.

zaro0508: only setting it once would be desirable but that can be in a follow on PR if you want

zaro0508: only setting it once would be desirable but that can be in a follow on PR if you want

tthyer: Agree will work on that next.

PR #45, opened 2020-01-29
Revert "Add manage policy to hide template to protect secrets"

zaro0508:
This reverts commit 9b2e4f3334b312c720e0fabe17c7754e48f026dd.
Revert this to debug permissions issue with S3 synapse bucket
provisiong using the SC

### DIFF
--- a/iam/sc-enduser-iam.yaml
+++ b/iam/sc-enduser-iam.yaml
@@ -27,18 +27,6 @@ Resources:
               AWS: !Sub 'arn:aws:iam::${AWS::AccountId}:root'
             Action:
               - 'sts:AssumeRole'
-  CFNDenyGetTemplateSummaryPolicy:
-    Type: 'AWS::IAM::ManagedPolicy'
-    Properties:
-      PolicyDocument:
-        Version: '2012-10-17'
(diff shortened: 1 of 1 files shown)


iam/sc-enduser-iam.yaml
```diff
@@ -27,18 +27,6 @@ Resources:
...
-    Properties:
-      PolicyDocument:
-        Version: '2012-10-17'
-        Statement:
-          -
-            Effect: Deny
-            Action:
-              # Deny this action to hide secrets
-              - cloudformation:GetTemplateSummary
-            Resource: "*"
```
tthyer: Why do you want to remove this entirely? It is not being applied to the group or role above. I anticipate we will want to re-apply this when we track down the problem.

zaro0508: this policy seems to apply to all resources?  I know it is deny me and I have nearly all permissions. https://pasteboard.co/ISfBKe3.png

tthyer: Ok let's give it a try. We will have to re-figure out how to hide our SSM secrets though

PR #53, opened 2020-02-01
Create development portfolio for EC2 with configurable group and role

tthyer:
New portfolio for EC2 development
Allows configuration of group and role

### DIFF
--- a/ec2/sc-portfolio-ec2-development.yaml
+++ b/ec2/sc-portfolio-ec2-development.yaml
@@ -1,0 +1,152 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Description: '[DEVELOPMENT] EC2 Portfolio for Service Catalog.'
+Metadata:
+  AWS::CloudFormation::Interface:
+    ParameterGroups:
+      - Label:
+          default: Portfolio Information
+        Parameters:
+          - PorfolioName
+          - PortfolioProvider
+          - PorfolioDescription
+      - Label:
+          default: IAM Settings
+        Parameters:
+          - LaunchRoleName
+          - PrincipalRoleName1
+          - PrincipalGroupName1
+          - PrincipalRoleName2
+          - PrincipalGroupName2
+      - Label:
+          default: Product Settings
+        Parameters:
+          - RepoRootURL
+Parameters:
+  PortfolioProvider:
+    Type: String
+    Description: Provider Name
+    Default: IT Services
+  PorfolioName:
+    Type: String
+    Description: Portfolio Name
+    Default: '[DEVELOPMENT] Service Catalog EC2 Reference Architecture'
+  PorfolioDescription:
+    Type: String
+    Description: Portfolio Description
+    Default: Service Catalog Portfolio that contains reference architecture
+      products for Amazon Elastic Compute Cloud (EC2).
+  LaunchRoleName:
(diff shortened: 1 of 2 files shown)


ec2/sc-portfolio-ec2-development.yaml
```diff
@@ -0,0 +1,152 @@
...
+    Type: String
+    Description: (Optional) Name of an IAM role which can execute products
+      in this portfolio.
+    Default: ''
+  PrincipalGroupName1:
+    Type: String
+    Description: (Optional) Name of an IAM  group which can execute products
+      in this portfolio.
+    Default: ''
+  PrincipalRoleName2:
```
zaro0508: why do we need a 2nd principal role? and why are both optional?  maybe update description or add a comment with explanation would help.

ec2/sc-portfolio-ec2-development.yaml
```diff
@@ -0,0 +1,152 @@
...
+      PortfolioId: !Ref 'SCEC2portfolio'
+      PrincipalType: IAM
+  LinkEndusersAltGroup:
+    Type: AWS::ServiceCatalog::PortfolioPrincipalAssociation
+    Condition: CondPrincipalGroupName2
+    Properties:
+      PrincipalARN: !Sub 'arn:aws:iam::${AWS::AccountId}:group/${PrincipalGroupName2}'
+      PortfolioId: !Ref 'SCEC2portfolio'
+      PrincipalType: IAM
+  ec2windowsjumpcloudproduct:
```
zaro0508: should we remove this same product from sc-portfolio-ec2.yaml file?

ec2/sc-portfolio-ec2-development.yaml
tthyer: Re alternate principal role/group, I was following the pattern that previously existed in the file where there were alternate roles. I can see that there might be a need for an alternate role/group in the future.

Why they are all optional: I don't want that to be the case, but which one am I going to require? The role, the group, or both?

ec2/sc-portfolio-ec2-development.yaml
tthyer: Yes, I can go ahead and do that on this PR. I do want to wrap this one up so that we have a pattern that I can apply to the S3 portfolios though.

ec2/sc-portfolio-ec2-development.yaml
zaro0508: it seems like sc-portfolio-ec2.yaml will create a user/role/group if one isn't specified and maybe that's why it's optional?   It seems like you'll need at least one group and role and the group and role needs to be created before executing this template.  That most likely means that we need a separate base template to create these groups and roles.  We would need to execute that base template before making this template available in the SC.  do we have such a base template?

tthyer:
@zaro0508 Responding to your question, yes, we have had such a base template all along, it is sc-enduser-iam. A few days ago you approved my [changes](https://github.com/Sage-Bionetworks/scipoolprod-sc-lib-infra/pull/52/files) to make the names of the group and role it creates configurable. Although it creates a group and a role with the same name, in the current PR I did not assume that sc-enduser-iam would be used, which means I do not assume that both a group and role exist, nor do I assume that they would be named the same.

PR #55, opened 2020-02-03
Update taskcat for EC2 and S3 portfolio template changes

tthyer:
Update taskcat for EC2 and S3 portfolio template changes


zaro0508:
Just run like travis to debug then create a new PR with the real fix.

PR #57, opened 2020-02-03
Fix S3 synapse bucket provisioning

zaro0508:
Our Synapse S3 bucket template contains a macro (lambda).  When provisioning
the template thru the SC it failed to execute the lambda because it didn't have
permissions.  It would fail with the following (not helpful) message:

  Failed to execute transform 237179673806::S3Objects. Rollback requested by user.

Setting the permissions in the lambda itself does not work because the service
catalog role will override those permissions.  Fix this by giving the SC role
the full lambda permissions.

### DIFF
--- a/iam/sc-s3-launchrole.yaml
+++ b/iam/sc-s3-launchrole.yaml
@@ -6,6 +6,7 @@ Resources:
       RoleName: SCS3LaunchRole
       ManagedPolicyArns:
         - arn:aws:iam::aws:policy/AmazonS3FullAccess
+        - arn:aws:iam::aws:policy/AWSLambdaFullAccess
       AssumeRolePolicyDocument:
         Version: 2012-10-17
         Statement:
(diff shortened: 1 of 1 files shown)


iam/sc-s3-launchrole.yaml
```diff
@@ -6,6 +6,7 @@ Resources:
       RoleName: SCS3LaunchRole
       ManagedPolicyArns:
         - arn:aws:iam::aws:policy/AmazonS3FullAccess
+        - arn:aws:iam::aws:policy/AWSLambdaFullAccess
```
tthyer: do you want to give the launch role full access to all lambdas rather than targeted access?

zaro0508: good question.  I think with the way that the SC is designed, in that just about any customization requires a lambda, I was thinking it might be general policy to give every SC role AWSLambdaFullAccess.  Or we can narrow down now and expand lambda access as we go? What do you think?

tthyer: I think we could try for more targeted access, if we can get that to work, as in [cfn-ssm-param-policy](https://github.com/Sage-Bionetworks/scipoolprod-infra/blob/master/templates/cfn-ssm-param-policy.yaml).

zaro0508: done

PR #60, opened 2020-02-03
Remove synapse product and add encrypted bucket product

tthyer:
products mixed up in s3-basic portfolio


tthyer:
@khai just swapping out these products.

PR #68, opened 2020-02-06
Always use private security group

tthyer:
- always use private security group
- if public network selected, additionally attach public security group

### DIFF
--- a/ec2/sc-ec2-linux-jumpcloud.yaml
+++ b/ec2/sc-ec2-linux-jumpcloud.yaml
@@ -235,7 +235,6 @@ Resources:
         - !Ref 'InstanceRole'
   PrivateSubnetSecurityGroup:
     Type: 'AWS::EC2::SecurityGroup'
-    Condition: UsePrivateSubnet
     Properties:
       GroupDescription: >-
         Allow all traffic to an instance originating from the VPN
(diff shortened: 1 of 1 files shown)


ec2/sc-ec2-linux-jumpcloud.yaml
```diff
@@ -235,7 +235,6 @@ Resources:
         - !Ref 'InstanceRole'
   PrivateSubnetSecurityGroup:
     Type: 'AWS::EC2::SecurityGroup'
-    Condition: UsePrivateSubnet
```
zaro0508: removing will always create this resource even if it's not being used?

tthyer: It will always be used, as with the templates in aws-infra

zaro0508: actually all ports are open for private instances so you don't even need a PrivateSecurityGroup.  The security group in managed-ec2-linux-v1.j2 is only created and used when the user specifies OpenPorts and it only applies to public instances

ec2/sc-ec2-linux-jumpcloud.yaml
```diff
@@ -436,15 +435,15 @@ Resources:
...
-          'Fn::Sub': '${AWS::Region}-internalpoolvpc-PrivateSubnet'
+        - UsePublicSubnet
         - !ImportValue
           'Fn::Sub': '${AWS::Region}-internalpoolvpc-PublicSubnet'
+        - !ImportValue
+          'Fn::Sub': '${AWS::Region}-internalpoolvpc-PrivateSubnet'
       SecurityGroupIds: !If
-        - UsePrivateSubnet
+        - UsePublicSubnet
+        - [!Ref PrivateSubnetSecurityGroup, !Ref PublicSubnetSecurityGroup]
```
zaro0508: why are both private subnet and public subnet here?

ec2/sc-ec2-linux-jumpcloud.yaml
zaro0508: sorry i misunderstood that your PrivateSubnetSecurityGroup is actually just the same as the Vpn security group.  this looks good.

ec2/sc-ec2-linux-jumpcloud.yaml
```diff
@@ -436,15 +435,15 @@ Resources:
...
         - [!Ref PrivateSubnetSecurityGroup]
```
zaro0508: why not just do 
```
- !ImportValue
              'Fn::Sub': '${AWS::Region}-${VpcName}-VpnSecurityGroup
```
like https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/managed-ec2-ubuntu-v1.yaml#L507

zaro0508: ohh yeah, reason is to avoid imports in libs

PR #70, opened 2020-02-07
Move windows product out of development portfolio

tthyer:
Move windows product out of development portfolio


tthyer:
@zaro0508 No one has tested Windows besides me. I kind of want someone who would actually use Windows to try it. The only testing I did was
1. does it create the instance
2. can I reach it with jumpcloud credentials
3. is it tagged

PR #71, opened 2020-02-11
SC-24

ahayden:
This statement should join strings to make a valid policy... but I'm not certain.

Passes pre-commit

### DIFF
--- a/s3/sc-s3-encrypted-ra.yaml
+++ b/s3/sc-s3-encrypted-ra.yaml
@@ -41,7 +41,7 @@ Resources:
             Sid: "ReadAccess"
             Effect: "Allow"
             Principal:
-              AWS: !Ref 'S3UserARNs'
+              - !Sub "AWS: [ ${S3UserARNs}, arn:aws:iam::325565585839:root ]"]
             Action:
               - "s3:ListBucket*"
               - "s3:GetBucketLocation"
--- a/s3/sc-s3-encrypted-ra.yaml
+++ b/s3/sc-s3-encrypted-ra.yaml
@@ -41,13 +41,7 @@ Resources:
             Sid: "ReadAccess"
             Effect: "Allow"
             Principal:
-              !Join
-                - ""
-                  - "AWS "
-                  - !Join
-                    - ","
-                      - !Ref ${S3UserARNs}
-                      - "arn:aws:iam::325565585839:root"
+              AWS: !Ref 'S3UserARNs'
             Action:
               - "s3:ListBucket*"
               - "s3:GetBucketLocation"
--- a/s3/sc-s3-encrypted-ra.yaml
+++ b/s3/sc-s3-encrypted-ra.yaml
@@ -41,7 +41,13 @@ Resources:
             Sid: "ReadAccess"
             Effect: "Allow"
             Principal:
-              - !Sub "AWS: [ ${S3UserARNs}, arn:aws:iam::325565585839:root ]"]
+              !Join
+                - ""
+                  - "AWS "
+                  - !Join
+                    - ","
+                      - !Ref ${S3UserARNs}
+                      - "arn:aws:iam::325565585839:root"
             Action:
               - "s3:ListBucket*"
               - "s3:GetBucketLocation"
--- a/s3/sc-s3-synapse-ra.yaml
+++ b/s3/sc-s3-synapse-ra.yaml
@@ -26,7 +26,7 @@ Parameters:
     Default: ""
 Conditions:
   HasBucketName: !Not [!Equals [!Ref BucketName, ""]]
-  AllowUserAccess: !Not [!Equals [!Join ['', !Ref S3UserARNs], "[]"]] #TODO investigate doesn't this not allow access to the principal?
+  AllowUserAccess: !Not [!Equals [!Join ['', !Ref S3UserARNs], "[]"]] 
 Resources:
   S3Bucket:
     Type: "AWS::S3::Bucket"
@@ -45,7 +45,7 @@ Resources:
             MaxAge: 3000
       Tags:
         - Key: Name
-          Value: SC-S3-RA-S3-Bucket #TODO should this be tagged with the Synapse userid at least
+          Value: SC-S3-RA-S3-Bucket 
   BucketPolicy:
     Type: "AWS::S3::BucketPolicy"
     Condition: AllowUserAccess
@@ -58,8 +58,13 @@ Resources:
             Sid: "ReadAccess"
             Effect: "Allow"
             Principal:
-              AWS: 
-              - !Sub "AWS: [ ${S3UserARNs}, arn:aws:iam::325565585839:root ]"] #Synapse-prod and requestor account read access
+              !Join
+                - ""
+                  - "AWS "
+                  - !Join
+                    - ","
+                      - !Ref ${S3UserARNs}
+                      - "arn:aws:iam::325565585839:root"
             Action:
               - "s3:ListBucket*"
               - "s3:GetBucketLocation"
--- a/s3/sc-s3-synapse-ra.yaml
+++ b/s3/sc-s3-synapse-ra.yaml
@@ -25,15 +25,15 @@ Parameters:
     Description: (Optional) Name of the created bucket.
     Default: ""
 Conditions:
-  HasBucketName: !Not [!Equals [!Ref BucketName, ""]]
+  HasSynapseUserName: !Not [!Equals [!Ref SynapseUserName, ""]]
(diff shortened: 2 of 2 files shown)


s3/sc-s3-encrypted-ra.yaml
```diff
@@ -41,7 +41,13 @@ Resources:
...
             Effect: "Allow"
             Principal:
-              AWS: !Ref 'S3UserARNs'
+              !Join
+                - ""
+                  - "AWS "
+                  - !Join
+                    - ","
+                      - !Ref ${S3UserARNs}
+                      - "arn:aws:iam::325565585839:root"
```
tthyer: SC-24 concerns the synapse bucket policy, not this bucket template, so please undo these changes

s3/sc-s3-synapse-ra.yaml
```diff
@@ -58,7 +58,13 @@ Resources:
...
             Effect: "Allow"
             Principal:
-              AWS: !Ref 'S3UserARNs'
+              !Join
+                - ""
+                  - "AWS "
+                  - !Join
+                    - ","
+                      - !Ref ${S3UserARNs}
+                      - "arn:aws:iam::325565585839:root"
```
tthyer: This looks odd. I don't think you need two joins. Don't need the `- "AWS"`. Does this template validate? I.e. if you run `aws cloudformation validate-template --template-body <path-to-your-file>`?

s3/sc-s3-synapse-ra.yaml
```diff
@@ -26,7 +26,7 @@ Parameters:
     Default: ""
 Conditions:
   HasBucketName: !Not [!Equals [!Ref BucketName, ""]]
-  AllowUserAccess: !Not [!Equals [!Join ['', !Ref S3UserARNs], "[]"]]
+  AllowUserAccess: !Not [!Equals [!Join ['', !Ref S3UserARNs], "[]"]]
```
zaro0508: undo this

s3/sc-s3-synapse-ra.yaml
```diff
@@ -45,7 +45,7 @@ Resources:
             MaxAge: 3000
       Tags:
         - Key: Name
-          Value: SC-S3-RA-S3-Bucket
+          Value: SC-S3-RA-S3-Bucket
```
zaro0508: undo this

s3/sc-s3-synapse-ra.yaml
```diff
@@ -67,7 +73,7 @@ Resources:
             Sid: "WriteAccess"
             Effect: "Allow"
             Principal:
-              AWS: !Ref 'S3UserARNs'
+              AWS: !Ref 'S3UserARNs' #sc-requestor write access
```
zaro0508: you'll need to do the same thing with `WriteAccess` as you did with `ReadAccess` because we want the ARN to have both read/list and write access.

s3/sc-s3-synapse-ra.yaml
tthyer: To be more clear, it currently looks like this:
```
Principal:
    AWS: !Ref 'S3UserARNs'
```
and I think whatever you're going to do should look something like this:
```
Principal:
    AWS: !Join ...
```
Join is going to create a string. Whatever is on the right side of "AWS:" needs to be a list. So it's possible that after combining the root ARN with the S3UserARNs list you'll have to Fn::Split.

zaro0508: maybe you can try this?
```
Principal: !Join
        - ','
        - 'arn:aws:iam::325565585839:root'
        - !Ref S3UserARNs
```

s3/sc-s3-synapse-ra.yaml
ahayden: This was one open question... Should the Synapse prod account really have write access to this bucket?

ahayden: This was one open question... Should the Synapse prod account really have write access to this bucket?

s3/sc-s3-synapse-ra.yaml
ahayden: The validation tool returns errors, but also errors with the file as it is in the master branch...

ahayden: This new commit validates OK

s3/sc-s3-synapse-ra.yaml
```diff
@@ -25,8 +25,9 @@ Parameters:
     Description: (Optional) Name of the created bucket.
     Default: ""
 Conditions:
-  HasBucketName: !Not [!Equals [!Ref BucketName, ""]]
+  HasSynapseUserName: !Not [!Equals [!Ref SynapseUserName, ""]]
```
zaro0508: This is a good thought but I don't think it's necessary.  The purpose of this template is to create a synapse bucket product so a SynapseUserName is required otherwise it won't be a synapse bucket.  

We should also remove the `default: ""` from properties section.

s3/sc-s3-synapse-ra.yaml
```diff
@@ -25,8 +25,9 @@ Parameters:
...
   AllowUserAccess: !Not [!Equals [!Join ['', !Ref S3UserARNs], "[]"]]
+  HasBucketName: !Not [!Equals [!Ref BucketName, ""]]
```
zaro0508: I don't see this being used anywhere?

s3/sc-s3-synapse-ra.yaml
```diff
@@ -58,7 +59,11 @@ Resources:
             Sid: "ReadAccess"
             Effect: "Allow"
             Principal:
-              AWS: !Ref 'S3UserARNs'
+              AWS: !Split
+                - ","
+                - !Sub
+                  - "${list},arn:aws:iam::325565585839:root"
```
zaro0508: would it be possible to parameterize the synapse ARN and then ref it here like so..
```
- "${list},${SynapsAccountArn}"
```
That will allow us to easily pass in a different ARN in case it changes for some reason.

s3/sc-s3-synapse-ra.yaml
```diff
@@ -58,7 +59,11 @@ Resources:
             Sid: "ReadAccess"
             Effect: "Allow"
             Principal:
-              AWS: !Ref 'S3UserARNs'
+              AWS: !Split
```
zaro0508: nice job finding this wacky yaml 👍

s3/sc-s3-synapse-ra.yaml
tthyer: So you want that to be a Parameter? I see your point, but that's another thing for the user to enter so opportunity for more user error. Could it be a Ref?

s3/sc-s3-synapse-ra.yaml
```diff
@@ -73,7 +72,11 @@ Resources:
             Sid: "WriteAccess"
             Effect: "Allow"
             Principal:
-              AWS: !Ref 'S3UserARNs' #sc-requestor write access
+              AWS: !Split
+                - ","
+                - !Sub
+                  - "${list},arn:aws:iam::325565585839:root" 
+                  - list: !Join [",",!Ref "S3UserARNs"]
```
tthyer: @ahayden I haven't ever seen this syntax before `list: ` -- do you have a reference for it? I can't find anything.

zaro0508: wacko! https://bl.ocks.org/magnetikonline/a6cfc522a1e9f876b75962f5f553c8e5

s3/sc-s3-synapse-ra.yaml
zaro0508: SC will always show the param even when there's a default?  can it be hidden with adjustments to `AWS::CloudFormation::Interface` section?

s3/sc-s3-synapse-ra.yaml
ahayden: fair enough... wanting to avoid future bugs

s3/sc-s3-synapse-ra.yaml
ahayden: I saw it in this gist... https://gist.github.com/magnetikonline/a6cfc522a1e9f876b75962f5f553c8e5

s3/sc-s3-synapse-ra.yaml
ahayden: Agreed- I'll define a simple parameter at the top that can be overridden from the default if needed

s3/sc-s3-synapse-ra.yaml
ahayden: This is checked a few lines later

PR #73, opened 2020-02-19
[SC-4] open application ports

zaro0508:
Allow SC users to open ports by selecting an application that will run
on the instance. The idea is to set ingress rules based on default ports
required by certain applications.  Deviating from the default application
ports would be a more custom setup which may require using the scicomp
provisioner


tthyer:
@zaro0508 This is an interesting idea, but we need to discuss how this fits with our other plans for offering other products.

PR #74, opened 2020-02-20
[IT-802] update AMI to ubuntu bionic v1.0.1 from imagecentral

zaro0508:


ec2/sc-ec2-linux-jumpcloud.yaml
```diff
@@ -30,7 +30,7 @@ Mappings:
     Ubuntu:
       AMIID: ami-07ebfd5b3428b6f4d
     AmazonLinux:
-      AMIID: ami-0ca7208f904dc1874
+      AMIID: ami-0944334364827dc2d
```
ahayden: Shouldn't this change be in the Ubuntu section?

zaro0508: woops! good catch.

PR #75, opened 2020-02-21
[SC-40] end user needs access to provision products

zaro0508:
We added deny GetTemplateSummary policy to restrict access to secrets however this also blocks provisioning of products.  Remove this policy to unblock provisioning.


zaro0508:
We haven't needed this so far so going to abandone it.

PR #77, opened 2020-02-25
WIP Sc 16/versioning demonstrated solution

tthyer:
Demonstrate a way to force the portfolio nested stack to always update. This requires that we use the sceptre-date-resolver in scipoolprod-infra.

### DIFF
--- a/deploy-dev-templates.sh
+++ b/deploy-dev-templates.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+set -ex
+
+# Deploy templates to AWS Admincentral account to share with other projects.
+REPO_NAME="${PWD##*/}"
+S3_BUCKET=$(aws --profile admincentral-cfservice cloudformation list-exports --query "Exports[?Name=='us-east-1-bootstrap-CloudformationBucket'].Value" --output text)
+S3_BUCKET_PATH="$REPO_NAME/$TRAVIS_BRANCH"
+S3_BUCKET_URL="s3://$S3_BUCKET/$S3_BUCKET_PATH"
+
+# Clean existing files on S3 bucket
+#aws s3 rm --recursive $S3_BUCKET_URL/
+
+# Upload dirs and files to S3 bucket
+# DIRS=$(ls -d */)
+# for dir in $DIRS
+# do
+#   aws s3 cp --recursive ${dir} $S3_BUCKET_URL/${dir}
+# done
+
+# Upload selected development templates
+S3_BUCKET_URL=s3://bootstrap-awss3cloudformationbucket-19qromfd235z9/scipoolprod-sc-lib-infra/master/
+templates=(ec2/sc-portfolio-ec2-development.yaml ec2/sc-ec2-linux-jumpcloud-v1.0.0.yaml ec2/sc-ec2-linux-jumpcloud-v1.0.2.yaml ec2/sc-product-ec2-linux-jumpcloud-development.yaml)
+#templates=(ec2/sc-product-ec2-linux-jumpcloud-development.yaml ec2/sc-ec2-linux-jumpcloud-v1.0.1.yaml)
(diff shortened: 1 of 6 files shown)


deploy-dev-templates.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
tthyer: Ignore this file -- was only used for testing.

ec2/sc-ec2-linux-jumpcloud-v1.0.0.yaml
```diff
@@ -0,0 +1,488 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
tthyer: Note file name. We should have a convention of adding a version number on the ends of the "LoadTemplateFromURL" templates.

ec2/sc-portfolio-ec2-development.yaml
```diff
@@ -126,6 +127,20 @@ Resources:
...
+    Properties:
+      Parameters:
+        PortfolioProvider: !Ref 'PortfolioProvider'
+        LaunchConstraintARN: !If
+          - CreateLaunchConstraint
+          - !GetAtt 'LaunchConstraintRole.Outputs.LaunchRoleArn'
+          - !Sub 'arn:aws:iam::${AWS::AccountId}:role/${LaunchRoleName}'
+        PortfolioId: !Ref 'SCEC2portfolio'
+        RepoRootURL: !Ref 'RepoRootURL'
+        StackDatetime: !Ref StackDatetime
```
tthyer: Note `StackDatetime` is passed from portfolio to product.

ec2/sc-product-ec2-linux-jumpcloud-development.yaml
```diff
@@ -0,0 +1,59 @@
...
+      SupportDescription: Sage IT
+      SupportEmail: it@sagebase.org
+      AcceptLanguage: en
+      SupportUrl: https://app.slack.com/client/T092YT0LV/CEFQD0KU1
+      ProvisioningArtifactParameters:
+        - Description: baseline version
+          Info:
+            LoadTemplateFromURL: !Sub '${RepoRootURL}ec2/sc-ec2-linux-jumpcloud-v1.0.0.yaml'
+          Name: v.1.0.0
+        - Description: !Sub 'Some minor update. Last updated ${StackDatetime}.'
```
tthyer: This is what triggers the update -- this text will change with every sceptre launch if the sceptre-date-resolver is used. Fortunately, this forces the template to be re-fetched.

ec2/sc-product-ec2-linux-jumpcloud-development.yaml
```diff
@@ -0,0 +1,59 @@
...
+      Distributor: !Ref 'PortfolioProvider'
+      SupportDescription: Sage IT
+      SupportEmail: it@sagebase.org
+      AcceptLanguage: en
+      SupportUrl: https://app.slack.com/client/T092YT0LV/CEFQD0KU1
+      ProvisioningArtifactParameters:
+        - Description: baseline version
+          Info:
+            LoadTemplateFromURL: !Sub '${RepoRootURL}ec2/sc-ec2-linux-jumpcloud-v1.0.0.yaml'
+          Name: v.1.0.0
```
zaro0508: version in `Name` and `LoadTemplateFromURL` values.  wondering if we can do a VAR for the version so don't have to repeat?

tthyer: No, I don't think so. We won't know how many versions there will be and would have to create new params for each. The only way this would work is if we could have a parameter which is a list of descriptions and a parameter which is a list of versions and then Explode them -- and explode isn't that sophisticated.

PR #78, opened 2020-02-25
SC-16 Add StackDatetime and allow product-level versioning

tthyer:
1. Add product-level versioning
2. Remove use of ProductVersionName used in portfolio-wide versioning
3. Add StackDatetime which will cause template to always be reloaded when used within ProvisioningArtifactParameters.

Doing this change in three parts:
1. Making changes as in the current pull request
2. Adding StackDatetime and removing ProductVersionName from the portfolio configuration in scipoolprod-infra
3. Creating an additional pull request here to remove ProductVersionName


tthyer:
@zaro0508 I will double-check I have the very latest commit. There is a companion PR for the S3 portfolio, #79. I had only planned to update portfolios that are actually in use.

ec2/sc-portfolio-ec2.yaml
```diff
@@ -137,7 +142,7 @@ Resources:
           - !Sub 'arn:aws:iam::${AWS::AccountId}:role/${LaunchRoleName}'
         PortfolioId: !Ref 'SCEC2portfolio'
         RepoRootURL: !Ref 'RepoRootURL'
-        ProductVersionName: !Ref 'ProductVersionName'
+        StackDatetime: !Ref StackDatetime
```
zaro0508: no longer using `ProductVersionName`?  maybe remove it from parameters list? you may need to update tests as well.

PR #82, opened 2020-03-02
SC-29 tag linux instance with name used for provisioned product

tthyer:
This adds a cfn-init script for tagging `Name` of a linux instance provisioned through Service Catalog, by looking up the name used for the provisioned product and using that for the tag value.

This PR depends upon https://github.com/Sage-Bionetworks/scipoolprod-infra/pull/67.

### DIFF
--- a/ec2/sc-ec2-linux-jumpcloud-v1.0.0.yaml
+++ b/ec2/sc-ec2-linux-jumpcloud-v1.0.0.yaml
@@ -227,6 +227,8 @@ Resources:
       ManagedPolicyArns:
         - !ImportValue
           'Fn::Sub': '${AWS::Region}-essentials-TagRootVolumePolicy'
+        - !ImportValue
+          'Fn::Sub': '${AWS::Region}-cfn-tag-instance-policy-TagInstancePolicy'
       AssumeRolePolicyDocument:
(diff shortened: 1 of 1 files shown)


ec2/sc-ec2-linux-jumpcloud-v1.0.0.yaml
```diff
@@ -436,6 +441,35 @@ Resources:
...
+        TagInstance:
+          files:
+            /opt/sage/bin/apply_name_tag.sh:
+              content: |
+                #! /bin/bash
+
+                set -xe
+
+                source /opt/sage/bin/instance_env_vars.sh
+                RESOURCE_ID=${STACK_ID##*/}
```
zaro0508: why you need `##*/`?

tthyer: I am trimming everything in the ARN that is not the part after the last forward-slash. I'm using bash native string manipulation to do so. If I try use the whole ARN it breaks the SearchQuery.

zaro0508: ahh ok.

PR #87, opened 2020-03-04
Create development-version products and templates for all existing products

tthyer:
Creates permanent development versions of products for all existing products. This allows faster development as the developer does not have to set up new products and edit the development portfolio over and over -- they only have to copy the latest version of the existing production product to the development version.

JIRA issue: https://sagebionetworks.jira.com/browse/SC-62


tthyer:
@zaro0508 I was trying to reduce the pita part!

PR #90, opened 2020-03-06
Add development notebook product

tthyer:
Adds product and template for a Linux notebook product.
Currently the only choice for "Notebook Type" is Rstudio. Eventually we'll add Jupyter as an option too.

Note, this is only the development version, I'll have to create a PR for the actual product too.


tthyer:
@zaro0508 This is only for the development product as I thought you would ask me to create a separate PR for the production version.

PR #95, opened 2020-03-18
[SC-70] remove OpenPorts parameter

zaro0508:
Only port that's allowed open when putting instance into the public
subnet is the SSH port.

### DIFF
--- a/ec2/sc-ec2-linux-jumpcloud-development.yaml
+++ b/ec2/sc-ec2-linux-jumpcloud-development.yaml
@@ -7,7 +7,6 @@ Metadata:
           default: General Configuration
         Parameters:
           - PrivateNetwork
-          - PrivateNetwork
       - Label:
           default: Linux Instance Configuration
         Parameters:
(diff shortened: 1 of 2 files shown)


ec2/sc-ec2-linux-jumpcloud-development.yaml
```diff
@@ -7,7 +7,7 @@ Metadata:
           default: General Configuration
         Parameters:
           - PrivateNetwork
-          - OpenPort
+          - PrivateNetwork
```
zaro0508: one to many of these

ec2/sc-ec2-linux-jumpcloud-v1.0.0.yaml
```diff
@@ -7,7 +7,7 @@ Metadata:
           default: General Configuration
         Parameters:
           - PrivateNetwork
-          - OpenPort
+          - PrivateNetwork
```
zaro0508: one too many

tthyer:
@zaro0508 you asked "do you think i need a new version of these files?" 
I think we should overwrite the current version. We don't want folks using the previous implementation because it is not as secure.

PR #96, opened 2020-03-19
Sc-83 change owner email source

tthyer:
1. Create a development folder for dev resources just to make the ec2 folder easier to manage
2. Used aws:servicecatalog:provisioningPrincipalArn to extract the owner's email. This is set into the environment variables and used in the script to register with Jumpcloud, the script to tag the root volume. Finally, I also apply OwnerEmail as a tag back to the instance.

follow up work:
1. do same for windows
2. move other development resources into development folder to tidy up ec2 dir

### DIFF
--- a/dev-deploy-templates.sh
+++ b/dev-deploy-templates.sh
@@ -11,9 +11,9 @@ S3_BUCKET_URL=s3://bootstrap-awss3cloudformationbucket-19qromfd235z9/scipoolprod
 
 # list the templates you're testing
 templates=(
-  ec2/sc-portfolio-ec2-development.yaml
-  ec2/sc-product-ec2-linux-jumpcloud-workflows-development.yaml
-  ec2/sc-ec2-linux-jumpcloud-workflows-development.yaml
+  ec2/development/sc-portfolio-ec2-development-tthyer.yaml
+  ec2/development/sc-product-ec2-linux-jumpcloud-development-tthyer.yaml
+  ec2/development/sc-ec2-linux-jumpcloud-development-tthyer.yaml
 )
 
 for template in "${templates[@]}"
--- a/ec2/sc-ec2-linux-jumpcloud-development.yaml
+++ b/ec2/sc-ec2-linux-jumpcloud-development.yaml
@@ -352,7 +352,8 @@ Resources:
                 }
                 DEPARTMENT=$(extract_tag_value Department)
                 PROJECT=$(extract_tag_value Project)
-                OWNER_EMAIL=$(extract_tag_value OwnerEmail)
+                PROVISIONING_PRINCIPAL_ARN=$(extract_tag_value 'aws:servicecatalog:provisioningPrincipalArn')
+                OWNER_EMAIL=${!PROVISIONING_PRINCIPAL_ARN##*/}
 
                 echo "export AWS_REGION=$AWS_REGION" > /opt/sage/bin/instance_env_vars.sh
                 echo "export STACK_NAME=$STACK_NAME" >> /opt/sage/bin/instance_env_vars.sh
@@ -445,7 +446,7 @@ Resources:
                 NAME=$(echo $PRODUCTS | jq '.ProvisionedProducts[0].Name')
                 /usr/bin/aws ec2 create-tags --region $AWS_REGION \
                   --resources $EC2_INSTANCE_ID \
-                  --tags Key=Name,Value=$NAME
+                  --tags Key=Name,Value=$NAME Key=OwnerEmail,Value=$OWNER_EMAIL
               mode: "00744"
               owner: "root"
               group: "root"
--- a/ec2/sc-ec2-linux-jumpcloud-notebook-development.yaml
+++ b/ec2/sc-ec2-linux-jumpcloud-notebook-development.yaml
@@ -339,7 +339,8 @@ Resources:
                 }
                 DEPARTMENT=$(extract_tag_value Department)
                 PROJECT=$(extract_tag_value Project)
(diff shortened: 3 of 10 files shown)


dev-deploy-templates.sh
```diff
@@ -11,9 +11,9 @@ S3_BUCKET_URL=s3://bootstrap-awss3cloudformationbucket-19qromfd235z9/scipoolprod
 
 # list the templates you're testing
 templates=(
-  ec2/sc-portfolio-ec2-development.yaml
-  ec2/sc-product-ec2-linux-jumpcloud-workflows-development.yaml
-  ec2/sc-ec2-linux-jumpcloud-workflows-development.yaml
+  ec2/development/sc-portfolio-ec2-development-tthyer.yaml
+  ec2/development/sc-product-ec2-linux-jumpcloud-development-tthyer.yaml
+  ec2/development/sc-ec2-linux-jumpcloud-development-tthyer.yaml
```
tthyer: creating a dev nested stack for myself to avoid collisions with other people doing development at same time

ec2/sc-ec2-linux-jumpcloud-development.yaml
```diff
@@ -352,7 +352,8 @@ Resources:
                 }
                 DEPARTMENT=$(extract_tag_value Department)
                 PROJECT=$(extract_tag_value Project)
-                OWNER_EMAIL=$(extract_tag_value OwnerEmail)
+                PROVISIONING_PRINCIPAL_ARN=$(extract_tag_value 'aws:servicecatalog:provisioningPrincipalArn')
+                OWNER_EMAIL=${!PROVISIONING_PRINCIPAL_ARN##*/}
```
tthyer: This is setting OWNER_EMAIL which goes into a file that can be source in different cfn-init steps to set values as environment vars. Since all that infrastructure is already in place, I don't need to update the other steps.

ec2/sc-ec2-linux-jumpcloud-development.yaml
```diff
@@ -445,7 +446,7 @@ Resources:
                 NAME=$(echo $PRODUCTS | jq '.ProvisionedProducts[0].Name')
                 /usr/bin/aws ec2 create-tags --region $AWS_REGION \
                   --resources $EC2_INSTANCE_ID \
-                  --tags Key=Name,Value=$NAME
+                  --tags Key=Name,Value=$NAME Key=OwnerEmail,Value=$OWNER_EMAIL
```
tthyer: @zaro0508 @ahayden Question: I went ahead and set `OwnerEmail` back on the instance (from the value derived from `aws:servicecatalog:provisioningPrincipalArn`). I was thinking this was a good idea as it's what we've been using. However, I think it's actually preferable to convert to using `aws:servicecatalog:provisioningPrincipalArn` as SC is going to tag that on all the resources it creates, except, evidently, EBS volumes. Thoughts?

ec2/sc-ec2-linux-jumpcloud-notebook-development.yaml
```diff
@@ -339,7 +339,8 @@ Resources:
                 }
                 DEPARTMENT=$(extract_tag_value Department)
                 PROJECT=$(extract_tag_value Project)
-                OWNER_EMAIL=$(extract_tag_value OwnerEmail)
+                PROVISIONING_PRINCIPAL_ARN=$(extract_tag_value 'aws:servicecatalog:provisioningPrincipalArn')
+                OWNER_EMAIL=${!PROVISIONING_PRINCIPAL_ARN##*/}
```
tthyer: Yes, I know I'm repeating this fix in a bunch of places. I want to try to unblock @ahayden for now and then I will backtrack and do SC-18, or SC-78 and SC-79.

ec2/sc-ec2-linux-jumpcloud-development.yaml
ahayden: Agreed- I think this is a good temporary fix. Maybe change the OwnerEmail tag to a UUID value in the future? This could be the Synapse ID of a user, which could also be the uid of a Jumpcloud user or the lookup value in the SSH pubkey table (when that happens).

ec2/sc-ec2-linux-jumpcloud-development.yaml
ahayden: I think you are correct that `aws:servicecatalog:provisioningPrincipalArn` is the input to use based on how we want SC to work via IAM users *and* assumed role with identity. Is the lack of tagging EBS a bug?

ec2/sc-ec2-linux-jumpcloud-development.yaml
zaro0508: ${!PROVISIONING_PRINCIPAL_ARN##*/} isn't going to give you the an email will it?  as is currently setup, won't it give you the synapse user id?

from..
https://docs.aws.amazon.com/cli/latest/reference/servicecatalog/search-provisioned-products.html

```
            "Tags": [
                {
                    "Value": "arn:aws:catalog:us-west-2:123456789012:product/prod-abcdfz3syn2rg",
                    "Key": "aws:servicecatalog:productArn"
                },
                {
                    "Value": "arn:aws:iam::123456789012:user/cliuser",
                    "Key": "aws:servicecatalog:provisioningPrincipalArn"
                },
```

tthyer: @zaro0508 Can we please fix this for current IAM users, which is the way we're supporting that people log in now, then we will do whatever retrofitting we need for the synapse solution when it is ready?

zaro0508: sure

PR #97, opened 2020-03-24
Reorganize dev templates and their deployment

tthyer:
- Move all "development" templates into development subfolders
- We've been moving toward something like this to allow for parallel development
- The script to deploy development templates now takes a file or directory argument


tthyer:
@zaro0508 @ahayden Let me know if you have a problem with these changes, and we can revisit. I'm going to merge so I'm not blocked.

PR #106, opened 2020-04-06
Add SSM service policy to instance templates

ahayden:
Passes pre-commit. 

This change adds the SSM core managed policy to the instance profiles of SC instances, similar to this: https://github.com/Sage-Bionetworks/aws-infra/pull/233


ec2/sc-ec2-linux-jumpcloud-notebook-v1.0.0.yaml
```diff
@@ -433,6 +435,10 @@ Resources:
           Value: !Sub "Service Catalog instance created by ${AWS::StackName}"
         - Key: "parkmycloud"
           Value: "yes"
+        - Key: "ManagedInstanceMaintenanceTarget"
```
zaro0508, line 438: could you please tell us what is the purpose of these tags?  how do you plan to use them?

Also should we apply these same tags to the templates in Sage-Bionetworks/aws-infra?

ec2/sc-ec2-linux-jumpcloud-v1.0.0.yaml
```diff
@@ -450,6 +452,10 @@ Resources:
           Value: !Sub "Service Catalog instance created by ${AWS::StackName}"
         - Key: "parkmycloud"
           Value: "yes"
+        - Key: "ManagedInstanceMaintenanceTarget"
+          Value: "yes"
+        - Key: "PatchGroup"
+          Value: "prod-default"
```
tthyer, line 458: I like this concept of different PatchGroups

ec2/sc-ec2-linux-jumpcloud-notebook-v1.0.0.yaml
ahayden: ManagedInstanceMaintenanceTarget is a boolean for whether we will manage it centrally or not. PatchGroup is a flag that the scheduled tasks can use to determine when and which patches to apply. For now, let's just use prod-default for SC, but later we may also want test-default, prod-updateall, prod-nopatch, etc.

PR #110, opened 2020-04-13
Add S3BucketTagger to s3 bucket templates

tthyer:
This adds the custom resource which tags S3 buckets to all the S3 templates in the library.
Depends on:
- https://github.com/Sage-Bionetworks-IT/cfn-cr-bucket-tagger/pull/1
- https://github.com/Sage-Bionetworks/scipoolprod-infra/pull/93

### DIFF
--- a/s3/development/sc-s3-encrypted-ra-development.yaml
+++ b/s3/development/sc-s3-encrypted-ra-development.yaml
@@ -77,7 +77,7 @@ Resources:
     Type: Custom::S3BucketTagger
     Properties:
       ServiceToken: !ImportValue
-        'Fn::Sub': '${AWS::Region}-set-bucket-tags-macro-SetBucketTagsFunctionArn'
+        'Fn::Sub': '${AWS::Region}-cfn-cr-bucket-tagger-FunctionArn'
       BucketName: !Ref S3Bucket
 Outputs:
   BucketName:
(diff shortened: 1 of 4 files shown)


s3/development/sc-s3-encrypted-ra-development.yaml
```diff
@@ -73,6 +73,13 @@ Resources:
             Condition:
               'Null':
                 s3:x-amz-server-side-encryption: 'true'
+  S3BucketTagger:
+    Type: Custom::S3BucketTagger
+    DependsOn: S3Bucket
```
zaro0508, line 78: i don't think you need this since there's already a !Ref on `BucketName`

s3/development/sc-s3-synapse-ra-development.yaml
```diff
@@ -100,6 +100,13 @@ Resources:
         ContentType: text
         ACL: authenticated-read
       Body: !Ref SynapseUserName
+  S3BucketTagger:
+    Type: Custom::S3BucketTagger
+    DependsOn: S3Bucket
```
zaro0508, line 105: same here

s3/development/sc-s3-encrypted-ra-development.yaml
tthyer: removed

s3/development/sc-s3-synapse-ra-development.yaml
tthyer: removed

PR #121, opened 2020-04-21
Fix RoleId import value to tag instances

ahayden:
This PR corrects the name of the ServiceCatalogEndusers role.

### DIFF
--- a/ec2/sc-ec2-linux-jumpcloud-v1.0.0.yaml
+++ b/ec2/sc-ec2-linux-jumpcloud-v1.0.0.yaml
@@ -442,7 +442,7 @@ Resources:
           Value: "prod-default"
           #The session name appended to this roleid becomes the value of the AccessApprovedCaller tag
         - Key: "AccessApprovedRoleId"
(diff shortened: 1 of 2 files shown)


ec2/sc-ec2-linux-jumpcloud-v1.0.0.yaml
```diff
@@ -442,7 +442,7 @@ Resources:
           Value: "prod-default"
           #The session name appended to this roleid becomes the value of the AccessApprovedCaller tag
         - Key: "AccessApprovedRoleId"
-          Value: !GetAtt 'ServiceCatalogEnduser.RoleId' 
+          Value: !GetAtt 'ServiceCatalogEndusers.RoleId'
```
zaro0508, line 445: There is no `ServiceCatalogEndusers` resource in this file.  You probably want an `!ImportValue` here?

ahayden: Ah right- I revised the PR to export it from the sc-enduser-iam resources and import it here.

PR #122, opened 2020-04-21
Remove RoleAndGroupName parameter from sc-enduser-iam

tthyer:
Removes parameter from enduser template that was causing complexity elsewhere. Should no longer be needed since we are splitting dev from prod account.

### DIFF
--- a/iam/sc-enduser-iam.yaml
+++ b/iam/sc-enduser-iam.yaml
@@ -1,14 +1,9 @@
 Description: "ServiceCatalog End User policy and group (fdp-1p4dlgcp7)"
-Parameters:
-  RoleAndGroupName:
-    Description: 'Name that will be used for the IAM Group and IAM Role'
-    Type: String
-    Default: ServiceCatalogEndusers
 Resources:
   SCEnduserGroup:
     Type: AWS::IAM::Group
     Properties:
-      GroupName: !Ref RoleAndGroupName
+      GroupName: ServiceCatalogEndusers
       ManagedPolicyArns:
         - arn:aws:iam::aws:policy/AWSServiceCatalogEndUserFullAccess
         - arn:aws:iam::aws:policy/AmazonEC2ReadOnlyAccess
@@ -23,7 +18,7 @@ Resources:
   SCEnduserRole:
     Type: AWS::IAM::Role
     Properties:
-      RoleName: !Ref RoleAndGroupName
+      RoleName: ServiceCatalogEndusers
       ManagedPolicyArns:
         - arn:aws:iam::aws:policy/AWSServiceCatalogEndUserFullAccess
         - !Ref SCEnduserExpandedPolicy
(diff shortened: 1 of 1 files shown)


iam/sc-enduser-iam.yaml
```diff
@@ -92,23 +92,23 @@ Outputs:
...
   EndUserGroupName:
     Value: !Ref SCEnduserGroup
     Export:
-      Name: !Sub '${AWS::Region}-${AWS::StackName}-${RoleAndGroupName}-GroupName'
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-ServiceCatalogEndusers-GroupName'
   EndUserRoleId:
     Value: !Ref SCEnduserRole.RoleId
     Export:
-      Name: !Sub '${AWS::Region}-${AWS::StackName}-${RoleAndGroupName}-RoleId'
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-ServiceCatalogEndusers-RoleId'
```
zaro0508, line 103: couldn't this just be `Name: !Sub '${AWS::Region}-${AWS::StackName}-RoleId`?

`StackName` makes it unique already so it will result in something like `us-east-1-sc-end-users-RoleId`, assuming sceptre we set the stack name to `sc-end-users` on deployment of sc-enduser-iam.yaml

tthyer: I think that would be a nice refinement that could be done in a follow-up PR. I am reluctant to go change this atm -- changing the group names leads to messy build failures because of their use in IAM accounts. I've had it work if you add the new group name in library, then change to the new name in infra, then remove the old name in the library.

iam/sc-enduser-iam.yaml
```diff
@@ -92,23 +87,23 @@ Outputs:
...
   EndUserRoleArn:
     Value: !GetAtt SCEnduserRole.Arn
     Export:
-      Name: !Sub '${AWS::Region}-${AWS::StackName}-${RoleAndGroupName}-RoleArn'
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-ServiceCatalogEndusers-RoleArn'
   EndUserRoleName:
     Value: !Ref SCEnduserRole
     Export:
-      Name: !Sub '${AWS::Region}-${AWS::StackName}-${RoleAndGroupName}-RoleName'
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-ServiceCatalogEndusers-RoleName'
```
zaro0508, line 106: should the ones below, `EndUserExpandPolicy` and `EndUserRemoteAccessPolicy` contain `ServiceCatalogEndusers` in export name as well?

iam/sc-enduser-iam.yaml
zaro0508: ok

iam/sc-enduser-iam.yaml
tthyer: No, those are for the policy resources. They are not exports that are from the role or group.

PR #127, opened 2020-04-23
Remove dev templates

tthyer:
Remove dev templates from library
This PR should be merged after https://github.com/Sage-Bionetworks/scipoolprod-infra/pull/99


tthyer:
@zaro0508 "this is gonna be a problem even with scipooldev" <-- what is "this"?

PR #136, opened 2020-04-24
[SC-127] Share tag optins template

zaro0508:
Move tag options template from scipoolprod-infra repo to this
repo so it can be shared accross multiple deployments.


zaro0508:
This move is a noop so merged without approval

PR #137, opened 2020-04-24
[SC-111] Refactor S3 portfolio template

zaro0508:
Refactor to use cross stack imports intead of passing in parameters.
We also removed unnessary paramters and make it more conducive for
sceptre deployment.

The main idea here is to make the template easier to understand for
preparation to remove 'AWS::CloudFormation::Stack' from this template

### DIFF
--- a/s3/sc-portfolio-s3-basic.yaml
+++ b/s3/sc-portfolio-s3-basic.yaml
@@ -1,26 +1,5 @@
 AWSTemplateFormatVersion: '2010-09-09'
 Description: S3 Portfolio for Service Catalog. (fdp-1p5rtpgl6)
-Metadata:
-  AWS::CloudFormation::Interface:
-    ParameterGroups:
-      - Label:
-          default: Portfolio Information
-        Parameters:
-          - PorfolioName
-          - PortfolioProvider
-          - PorfolioDescription
-      - Label:
-          default: IAM Settings
-        Parameters:
-          - LaunchRoleName
-          - PrincipalRoleName1
-          - PrincipalGroupName1
-          - PrincipalRoleName2
-          - PrincipalGroupName2
-      - Label:
-          default: Product Settings
-        Parameters:
-          - RepoRootURL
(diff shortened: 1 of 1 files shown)


s3/sc-portfolio-s3-basic.yaml
```diff
@@ -66,76 +22,34 @@ Parameters:
...
       Description: !Ref 'PorfolioDescription'
       DisplayName: !Ref 'PorfolioName'
-  LaunchConstraintRole:
-    Type: AWS::CloudFormation::Stack
-    Condition: CreateLaunchConstraint
-    Properties:
-      TemplateURL: !Sub '${RepoRootURL}iam/sc-s3-launchrole.yaml'
-      TimeoutInMinutes: 5
   LinkEndusersRole:
     Type: AWS::ServiceCatalog::PortfolioPrincipalAssociation
```
ahayden, line 33: How about leaving the condition statements the restrict which roles can execute products? We might use these conditions if we start differentiating roles

s3/sc-portfolio-s3-basic.yaml
```diff
@@ -66,76 +22,34 @@ Parameters:
...
-    Condition: CondPrincipalRoleName1
     Properties:
-      PrincipalARN: !Sub 'arn:aws:iam::${AWS::AccountId}:role/${PrincipalRoleName1}'
+      PrincipalARN: !ImportValue
+        'Fn::Sub': '${AWS::Region}-sc-enduser-iam-ServiceCatalogEndusers-RoleArn'
       PortfolioId: !Ref 'SCS3portfolio'
       PrincipalType: IAM
   LinkEndusersGroup:
     Type: AWS::ServiceCatalog::PortfolioPrincipalAssociation
-    Condition: CondPrincipalGroupName1
```
ahayden, line 111: Consider leaving this condition too, which would give us the option of grouping roles. Maybe we want a Synapse team to map to a group?

PR #146, opened 2020-04-29
Allow reading role information in Linux instance profile

ahayden:
This change allows an instance to read the roleid of a rolename in order to write a tag to itself. This permission set should be limited to Service Catalog instances that need to limit access to OIDC sessionNames. Thank you, @tthyer for raising the issue of instance profiles.


zaro0508:
So have you confirmed that modifying an instance profile will not rebuild an instance?

ahayden:
I haven't confirmed that yet. This will be the test. I don't believe and prod instances will be effected.

PR #151, opened 2020-04-30
Use bucket policy custom resource in s3 bucket templates

tthyer:
Add bucket policy custom resource to s3 templates.
This custom resource will create a policy that adds the provisioning user as a principal, allows extra principals to be added, and optionally requires encryption.
This PR depends on https://github.com/Sage-Bionetworks/scipool-infra/pull/115
For more information about the custom resource, see https://github.com/Sage-Bionetworks-IT/cfn-cr-sc-bucket-policy/pull/1

### DIFF
--- a/s3/sc-s3-encrypted-ra-v1.0.0.yaml
+++ b/s3/sc-s3-encrypted-ra-v1.0.0.yaml
@@ -1,9 +1,14 @@
 AWSTemplateFormatVersion: '2010-09-09'
-Description: 'Service Catalog: S3 Reference Architecture: Private encrypted bucket
-  using AES256 with retention on object deletion. (fdp-1oc5gsre6)'
+Description: >-
+  Service Catalog S3 Reference Architecture: Private encrypted bucket
+  using AES256 with retention on object deletion.
 Metadata:
   AWS::CloudFormation::Interface:
     ParameterGroups:
+      - Label:
+          default: S3 Bucket Configuration
+        Parameters:
+          - BucketName
       - Label:
           default: S3 Bucket Policy Configuration
         Parameters:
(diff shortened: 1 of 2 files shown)


s3/sc-s3-encrypted-ra-v1.0.0.yaml
```diff
@@ -1,84 +1,55 @@
...
-            Condition:
-              'Null':
-                s3:x-amz-server-side-encryption: 'true'
   S3BucketTagger:
     Type: Custom::S3BucketTagger
     Properties:
       ServiceToken: !ImportValue
         'Fn::Sub': '${AWS::Region}-cfn-cr-bucket-tagger-FunctionArn'
       BucketName: !Ref S3Bucket
+  S3BucketPolicy:
```
zaro0508, line 45: Does this depend on `S3BucketTagger`?  if so we may need a `DependsOn` here?

s3/sc-s3-synapse-ra-v1.0.0.yaml
```diff
@@ -47,43 +54,17 @@ Resources:
...
-                - !Sub
-                  - "${list},${SynapseProdARN}" 
-                  - list: !Join [",",!Ref "S3UserARNs"]
-            Action:
-              - "s3:*Object*"
-              - "s3:*MultipartUpload*"
-            Resource: !Sub "${S3Bucket.Arn}/*"
+      ServiceToken: !ImportValue
+        'Fn::Sub': '${AWS::Region}-cfn-cr-sc-bucket-policy-FunctionArn'
+      BucketName: !Ref S3Bucket
```
zaro0508, line 62: don't we need to set `RequireEncryption` to true here since BucketEncryption is enabled?

s3/sc-s3-encrypted-ra-v1.0.0.yaml
tthyer: They are completely independent.

s3/sc-s3-synapse-ra-v1.0.0.yaml
tthyer: The current Synapse bucket does not in fact use the Deny statements. I am reproducing what currently exists in the bucket templates.

zaro0508: I still think that it makes more sense to match `BucketEncryption enabled` with `RequireEncryption=true` even though synapse documentation doesn't require it.  I'm guessing it's a bug in the synapse docs.

In any case, i'm not opposed to setting to false either since synapse docs don't mention it currently.

PR #156, opened 2020-05-13
replace ConnectionInstructions with  ConnectionURI

zaro0508:
Now that we have ConnectionURI to allow a direct SSH session
manager connection we no longer need the ConnectionInstructions

### DIFF
--- a/ec2/sc-ec2-linux-jumpcloud-notebook-v1.0.0.yaml
+++ b/ec2/sc-ec2-linux-jumpcloud-notebook-v1.0.0.yaml
@@ -450,5 +450,5 @@ Outputs:
     Value: !Ref 'InstanceRole'
   IAMInstanceProfile:
     Value: !Ref 'InstanceProfile'
-  ConnectionInstructions:
-    Value: "https://sagebionetworks.jira.com/wiki/spaces/IT/pages/996376579/Connect+to+Provisioned+Instances"
+  ConnectionURI:
+    Value: !Sub "https://${AWS::Region}.console.aws.amazon.com/systems-manager/session-manager/${LinuxInstance}?region=${AWS::Region}"
--- a/ec2/sc-ec2-windows-jumpcloud-v1.0.0.yaml
+++ b/ec2/sc-ec2-windows-jumpcloud-v1.0.0.yaml
@@ -274,5 +274,5 @@ Outputs:
     Value: !Ref 'InstanceRole'
   IAMInstanceProfile:
     Value: !Ref 'InstanceProfile'
-  ConnectionInstructions:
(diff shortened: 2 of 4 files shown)


ec2/sc-ec2-windows-jumpcloud-v1.0.0.yaml
```diff
@@ -274,5 +274,5 @@ Outputs:
     Value: !Ref 'InstanceRole'
   IAMInstanceProfile:
     Value: !Ref 'InstanceProfile'
-  ConnectionInstructions:
-    Value: "https://sagebionetworks.jira.com/wiki/spaces/IT/pages/996376579/Connect+to+Provisioned+Instances"
+  ConnectionURI:
```
ahayden, line 277: This technically *can* work for Windows hosts, but it will be a Powershell session. We also haven't yet set up the correct AccessApprovedCaller tag for Windows... that needs to happen in a script that runs when the instance first boots.

ec2/sc-ec2-linux-jumpcloud-notebook-v1.0.0.yaml
```diff
@@ -450,5 +450,5 @@ Outputs:
     Value: !Ref 'InstanceRole'
   IAMInstanceProfile:
     Value: !Ref 'InstanceProfile'
-  ConnectionInstructions:
```
ahayden, line 453: In my opinion, we should leave the connection instructions output with some explanation of how to connect, as the ConnectionURI is not self-explanatory in all cases. 

For Notebooks, for example, we will change the ConnectionURI to an ALB DNS record at some point, for for our first release it will be an internal IP addr, and we need a way to point people to a VPN route and to which credentials to use.

zaro0508: How about add it back in when you have the docs for it?

ec2/sc-ec2-windows-jumpcloud-v1.0.0.yaml
zaro0508: ok, will remove for windows.  can add it when, and if, you decide to enable powershell access

ec2/sc-ec2-linux-jumpcloud-notebook-v1.0.0.yaml
tthyer: It is ok to remove ConnectionInstructions from the notebook template iff you add a separate property for starting a browser connection to the notebook server. You could name it something like `NotebookServerUri`, as discussed at standup.

zaro0508:
replacing this PR with PR https://github.com/Sage-Bionetworks/service-catalog-library/pull/174

PR #158, opened 2020-05-13
SC-138 Add shell access to workflows instances and use protected path in enduser role access policy

ahayden:
This change will cause the enduser role to allow access by looking at the Protected/AccessApprovedCaller tag on an instance. This tagged is added by an instance when it boots and runs the service-catalog-utils scripts.

This change also adds the appropriate tags to workflows EC2s.


ec2/sc-ec2-linux-jumpcloud-workflows-v1.0.0.yaml
```diff
@@ -448,3 +452,5 @@ Outputs:
     Value: !Ref 'InstanceProfile'
   ConnectionInstructions:
     Value: "https://sagebionetworks.jira.com/wiki/spaces/IT/pages/996376579/Connect+to+Provisioned+Instances"
+  ConnectionURI:
```
zaro0508, line 455: i believe this is already done in PR https://github.com/Sage-Bionetworks/service-catalog-library/pull/156

ec2/sc-ec2-linux-jumpcloud-workflows-v1.0.0.yaml
```diff
@@ -429,6 +429,10 @@ Resources:
           Value: "yes"
         - Key: "PatchGroup"
           Value: "prod-default"
+          #The session name appended to this roleid becomes the value of the AccessApprovedCaller tag
+        - Key: "AccessApprovedRoleId"
```
zaro0508, line 433: why not add this to the other EC2 templates as well?

tthyer: same question -- if we know this works, we need edits on the other product templates

ec2/sc-ec2-linux-jumpcloud-workflows-v1.0.0.yaml
ahayden: agreed- let's leave those changes to PR #156

ec2/sc-ec2-linux-jumpcloud-workflows-v1.0.0.yaml
ahayden: There's no harm in adding it to all the templates as long as access methods are confusing to endusers. Maybe people will want to shell into their notebooks. Let's put the tag on Windows instances too and at some point we can change the boot scripts to apply the Protected/AccessApprovedCaller tag if we think it would be useful.

PR #159, opened 2020-05-13
[SC-137] Fix jumpcloud integration

zaro0508:
Created `SynapseEmail` attribute in jumpcloud and make instance
initilization use that value instead of the jumpcloud `email` value
to associate JC instances to users.

The SynapseEmail attribute would have a value like "jsmith@synapse.org"

I believe this also fixes issue SC-145

The windows fix depends on PR https://github.com/Sage-Bionetworks/infra-utils/pull/45
and will require a new infra-utils tag


zaro0508:
fix for windows added @tthyer, it depends on the PR in the commit message.

PR #168, opened 2020-05-21
SC-143: Add trust relationship with synapse-login instance role

brucehoff:
Add trust relationship with synapse-login instance role

### DIFF
--- a/iam/sc-enduser-iam.yaml
+++ b/iam/sc-enduser-iam.yaml
@@ -30,12 +30,10 @@ Resources:
         Statement:
           - Effect: Allow
             Principal:
-              AWS: !Sub 'arn:aws:iam::${AWS::AccountId}:root'
-              # Synapse OIDC identity provider
-              Federated: !Sub 'arn:aws:iam::${AWS::AccountId}:oidc-provider/repo-prod.prod.sagebase.org/auth/v1'
+              AWS: !Sub 'arn:aws:iam::${AWS::AccountId}:role/synapse-login-scipooldev-InstanceRole-*'
             Action:
-              - 'sts:AssumeRoleWithWebIdentity'
               - 'sts:AssumeRole'
+              - 'sts:TagSession'
   SCEnduserExpandedPolicy:
     Type: AWS::IAM::ManagedPolicy
     Properties:
--- a/iam/sc-enduser-iam.yaml
+++ b/iam/sc-enduser-iam.yaml
@@ -30,7 +30,7 @@ Resources:
         Statement:
           - Effect: Allow
             Principal:
-              AWS: !Sub 'arn:aws:iam::${AWS::AccountId}:role/synapse-login-scipooldev-InstanceRole-*'
+              AWS: !Sub 'arn:aws:iam::${AWS::AccountId}:role/synapse-login-scipool*-InstanceRole-*'
             Action:
               - 'sts:AssumeRole'
               - 'sts:TagSession'


iam/sc-enduser-iam.yaml
```diff
@@ -30,12 +30,10 @@ Resources:
         Statement:
           - Effect: Allow
             Principal:
-              AWS: !Sub 'arn:aws:iam::${AWS::AccountId}:root'
-              # Synapse OIDC identity provider
-              Federated: !Sub 'arn:aws:iam::${AWS::AccountId}:oidc-provider/repo-prod.prod.sagebase.org/auth/v1'
+              AWS: !Sub 'arn:aws:iam::${AWS::AccountId}:role/synapse-login-scipool*-InstanceRole-*'
```
zaro0508, line 33: if this works then i'm fine with it.  Just wondering if this is a use case for using the `Condition` property instead of `*`?  https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-idp_oidc.html

brucehoff: Are you suggesting using "StringLike" rather than *?  I'm happy to do so if that's a more standard way to filter.  

(Note that the page you linked is for connecting an OIDC provider, which we are no longer doing.)

iam/sc-enduser-iam.yaml
```diff
@@ -30,12 +30,10 @@ Resources:
         Statement:
           - Effect: Allow
             Principal:
-              AWS: !Sub 'arn:aws:iam::${AWS::AccountId}:root'
-              # Synapse OIDC identity provider
-              Federated: !Sub 'arn:aws:iam::${AWS::AccountId}:oidc-provider/repo-prod.prod.sagebase.org/auth/v1'
```
zaro0508, line 35: I think removing this breaks the ConnectionURI, the ability to do an SSH connection using the AWS web console.  The error when clicking on the `ConnectionURI` link is: 
```
User: arn:aws:sts::465877038949:assumed-role/ServiceCatalogEndusers/3377358 is not authorized to perform: ssm:StartSession on resource: arn:aws:ec2:us-east-1:465877038949:instance/i-02b1c202c36e86c93
```
It causes the `EndUserRoleId` to be null at https://github.com/Sage-Bionetworks/service-catalog-library/blob/master/iam/sc-enduser-iam.yaml#L102 

The EC2 templates rely on the `EndUserRoleId` info here https://github.com/Sage-Bionetworks/service-catalog-library/blob/master/ec2/sc-ec2-linux-jumpcloud-v1.0.0.yaml#L430

brucehoff: I checked scipooldev which has a manually implemented version of this change, and I *was* able to use a ConnectionUri.

iam/sc-enduser-iam.yaml
zaro0508: i guess i like the way that works.

brucehoff: I just realized I should try this "manually" in scipool dev before we merge and deploy the PR.  The result was:
```
An error occurred: Invalid principal in policy: "AWS":"arn:aws:iam::465877038949:role/synapse-login-scipool*-InstanceRole-*"
```
`"arn:aws:iam::465877038949:role/*"` didn't work either:
```
An error occurred: Invalid principal in policy: "AWS":"arn:aws:iam::465877038949:role/*"
```
So perhaps wildcards are not allowed.

brucehoff: I will close this PR and open a new one once I verify it works via manual deployment.

PR #186, opened 2020-05-27
Replace support with products snippet

zaro0508:
PR https://github.com/Sage-Bionetworks/admincentral-infra/pull/84
created a new snippet (products.yaml).  The new snippet adds
owner and Distributor properties.  Now we can use the products
snippet for all Sage products.

### DIFF
--- a/ec2/sc-product-ec2-linux-jumpcloud-notebook.yaml
+++ b/ec2/sc-product-ec2-linux-jumpcloud-notebook.yaml
@@ -1,10 +1,6 @@
 AWSTemplateFormatVersion: '2010-09-09'
 Description: 'Notebook Linux EC2 ServiceCatalog product with Jumpcloud integration.'
 Parameters:
-  PortfolioProvider:
-    Type: String
-    Description: Owner and Distributor Name
-    Default: IT Services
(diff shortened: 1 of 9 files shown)


ec2/sc-product-ec2-linux-jumpcloud-notebook.yaml
```diff
@@ -41,6 +37,7 @@ Resources:
       'Fn::Transform':
         Name: 'AWS::Include'
         Parameters:
+          # source: https://github.com/Sage-Bionetworks/admincentral-infra/blob/master/templates/cfn-snippets-bucket.yaml
```
tthyer, line 40: Why comment out rather than remove?

zaro0508: this is not a CFN property, although i can see why you would think it is.    it's just a comment to indicate the source of the snippet otherwise we may not remember where to look to make updates.

tthyer: Oh, I see. Thanks for the clarification.

PR #188, opened 2020-05-27
Output public IP address

zaro0508:
The EC2 products templates support placing instances in public subnets
however it neglects to provide the public IP info.  This will display
the public IP address to end users if they choose to provision into
a public subnet.


zaro0508:
did you want to abandone this PR or just wait for a decision to SC-167 before merging @tthyer ?

zaro0508:
We have decided to drop support for for public subnet.

PR #195, opened 2020-06-08
[SC-166] Remove SynapseProdARN

zaro0508:
The synapse AWS account is very unlikely to change therefore
no need to parameterize SynapseProdARN

### DIFF
--- a/s3/sc-s3-synapse-ra-v1.0.1.yaml
+++ b/s3/sc-s3-synapse-ra-v1.0.1.yaml
@@ -1,0 +1,98 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Transform: S3Objects
+Description: >-
+  Synapse S3 Custom Storage
+  (https://docs.synapse.org/articles/custom_storage_location.html)
+Metadata:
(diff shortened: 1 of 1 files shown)


s3/sc-s3-synapse-ra-v1.0.1.yaml
```diff
@@ -0,0 +1,98 @@
...
+          default: S3 Bucket Configuration
+        Parameters:
+          - BucketName
+      - Label:
+          default: S3 Bucket Policy Configuration
+        Parameters:
+          - SynapseUserName
+          - S3UserARNs
+Parameters:
+  SynapseUserName:
```
xschildw, line 19: Supported but the recommended value is now Synapse userId or teamId.
https://docs.synapse.org/articles/custom_storage_location.html

zaro0508: That's a separate issue.  Would you mind creating a jira issue and we can address it in a separate PR.

xschildw: Sure.

PR #197, opened 2020-06-09
[SC-39] Remove SCEnduserGroup

zaro0508:
The SCEnduserGroup was initially setup for IAM users to access the SC.
Now that we have users loging in thru synapse and assuming the
SCEnduserRole we no longer need SCEnduserGroup.

Note: deploying this PR requires that we first remove the `LinkEndusersGroup`
from the S3 and EC2 portfolios.


zaro0508:
I have tested this on scipooldev so i think we are good to go.

PR #198, opened 2020-06-09
[SC-100] Remove jumpcloud

zaro0508:
Only support Connection URI, no more jumpcloud access.


zaro0508:
I've removed the below section from the v1.0.0 file.  I'm not sure what it does or wether it should be replaced with something else so i removed.

```
        add_docker_user:
          commands:
            01_add_jc_user_to_docker_group:
              command: "gpasswd -a $(cat /opt/sage/jcusername) docker"
```

zaro0508:
We have decided to not merge this until until we figure out a way to increase the timeout with the ConnectionURI (issue SC-160).

zaro0508:
replace this PR with #216

PR #203, opened 2020-06-17
[SC-65] External EC2 products

zaro0508:
Create a EC2 and S3 portfolios for external (non sage) users
* New IAM role for external users
* New portofolio for external users
* Add AWS linux and Ubuntu instance to EC2 product


s3/sc-portfolio-s3-basic-external.yaml
```diff
@@ -0,0 +1,34 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
tthyer, line 1: We were not planning to offer an S3 portfolio to external users

iam/sc-enduser-iam.yaml
```diff
@@ -22,6 +22,28 @@ Resources:
...
+        Statement:
+          - Effect: Allow
+            Principal:
+              AWS: "*"
+            Condition:
+              StringLike:
+                aws:PrincipalArn: !Sub 'arn:aws:iam::${AWS::AccountId}:role/synapse-login-scipool*-InstanceRole-*'
+            Action:
+              - 'sts:AssumeRole'
+              - 'sts:TagSession'
```
tthyer, line 46: Why is there a separate end user role? Isn't this going to require that we change login app (a lot)? We were planning to use  ABAC with session tags to control portfolio access.

zaro0508: it's because portfolio access is done thru roles and we need another role to associate external portfolios to, https://github.com/Sage-Bionetworks/service-catalog-library/blob/master/ec2/sc-portfolio-ec2.yaml#L26

So there is no ABAC usage setup yet. Do you have a suggestion as to what should be done now?  Setup  ABAC  before creating external portfolio? Setup ABAC as a follow on PR?  I'm open to any suggestion you want to propose.

s3/sc-portfolio-s3-basic-external.yaml
zaro0508: Ok. will remove external S3 portfolio

PR #213, opened 2020-06-26
Add import security group for alb ingress and new boot script

ahayden:
This PR adds a security group to the notebook instance type that allowing ingress from the ALB in the appropriate Service Catalog account. 

The notebook template now includes a reference to a boot script in the service-catalog-utils repo that will substitute a variable into a path in the apache proxy config file. This script will only be needed for instances that run internal web servers and have the apache proxy, and so isn't needed for other instance types.

depends on https://github.com/Sage-Bionetworks/service-catalog-utils/pull/10  https://github.com/Sage-Bionetworks/service-catalog-library/pull/213 https://github.com/Sage-Bionetworks/scipool-infra/pull/143 https://github.com/Sage-Bionetworks/service-catalog-utils/pull/11

### DIFF
--- a/ec2/sc-ec2-linux-jumpcloud-notebook.yaml
+++ b/ec2/sc-ec2-linux-jumpcloud-notebook.yaml
@@ -266,7 +266,7 @@ Resources:
                 [cfn-auto-reloader-hook]
                 triggers=post.update
                 path=Resources.LinuxInstance.Metadata.AWS::CloudFormation::Init
-                action=/opt/aws/bin/cfn-init -v --stack ${AWS::StackName} --resource LinuxInstance --configsets SetupCfn,SetEnv,SetTags,SetupJumpcloud --region ${AWS::Region}
+                action=/opt/aws/bin/cfn-init -v --stack ${AWS::StackName} --resource LinuxInstance --configsets SetupCfn,SetEnv,SetTags,SetupApacheConf,SetupJumpcloud --region ${AWS::Region}
               mode: "000400"
               owner: root
               group: root
@@ -377,7 +377,7 @@ Resources:
       UserData:
         Fn::Base64: !Sub |
           #!/bin/bash
-          /opt/aws/bin/cfn-init -v --stack ${AWS::StackName} --resource LinuxInstance --configsets SetupCfn,SetEnv,SetTags,SetupJumpcloud --region ${AWS::Region}
+          /opt/aws/bin/cfn-init -v --stack ${AWS::StackName} --resource LinuxInstance --configsets SetupCfn,SetEnv,SetTags,SetupApacheConf,SetupJumpcloud --region ${AWS::Region}
           /opt/aws/bin/cfn-signal -e $? --stack ${AWS::StackName} --resource LinuxInstance --region ${AWS::Region}
       Tags:
         - Key: Name
--- a/ec2/sc-ec2-linux-jumpcloud-notebook.yaml
+++ b/ec2/sc-ec2-linux-jumpcloud-notebook.yaml
@@ -247,7 +247,7 @@ Resources:
           SetupJumpcloud:
             - install_jc
             - config_jc
-          WriteApacheConf:
+          SetupApacheProxy:
             - WriteApacheConf
         cfn_hup_service:
           files:
--- a/ec2/sc-ec2-linux-jumpcloud-notebook.yaml
+++ b/ec2/sc-ec2-linux-jumpcloud-notebook.yaml
@@ -266,7 +266,7 @@ Resources:
                 [cfn-auto-reloader-hook]
                 triggers=post.update
                 path=Resources.LinuxInstance.Metadata.AWS::CloudFormation::Init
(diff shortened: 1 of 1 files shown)


zaro0508:
It would help if you could reference all the dependencies with a "depend on <PR>" in the PR description.

ec2/sc-ec2-linux-jumpcloud-notebook.yaml
```diff
@@ -19,7 +19,7 @@ Metadata:
 Mappings:
   NotebookTypes:
     Rstudio:
-      AMIID: "ami-0ed2999f7cff2e3e3"  # https://github.com/Sage-Bionetworks-IT/packer-rstudio/tree/v1.0.7
+      AMIID: "ami-051310d409a32a5aa"  # https://github.com/Sage-Bionetworks-IT/packer-rstudio/tree/v1.0.7
```
tthyer, line 22: @ahayden The AMIID looks right but the comment there should say v2.0.0 no v1.0.7

ec2/sc-ec2-linux-jumpcloud-notebook.yaml
```diff
@@ -326,20 +326,31 @@ Resources:
...
+        WriteApacheConf:
+          files:
+            /opt/sage/bin/apache_conf_rewrite.sh:
+              source: "https://raw.githubusercontent.com/Sage-Bionetworks/service-catalog-utils/v1.0.6/linux/opt/sage/bin/apache_conf_rewrite.sh"
+              mode: "00744"
+              owner: "root"
+              group: "root"
+          commands:
+            01_rewrite_apache_conf:
+              command: "/bin/bash /opt/sage/bin/apache_conf_rewrite.sh"
```
tthyer, line 345: Note that you need to do more than just defining this `WriteApacheConf` block. You need to actually get things to use it. Notice how `TagInstance` defined above is used: it is one of the steps included in the `SetTags` configSet. That configSet is passed to the cfn-init script as a parameter in the contents of the file cfn-auto-reloader.conf and in `UserData`.

ahayden: OK, interesting... I did not know.. I'll add a new configSet.

ec2/sc-ec2-linux-jumpcloud-notebook.yaml
```diff
@@ -326,20 +328,31 @@ Resources:
...
     Properties:
       ImageId: !FindInMap [NotebookTypes, !Ref NotebookType, AMIID]
       InstanceType: !Ref 'EC2InstanceType'
       SubnetId: !ImportValue
         'Fn::Sub': '${AWS::Region}-${VpcName}-PrivateSubnet'
       SecurityGroupIds:
         - 'Fn::ImportValue': !Sub '${AWS::Region}-${VpcName}-VpnSecurityGroup'
+        - 'Fn::ImportValue': !Sub '${AWS::Region}-alb-notebook-access-EC2SecurityGroup'
```
zaro0508, line 355: As mentioned in https://github.com/Sage-Bionetworks/scipool-infra/pull/143/files#r450933540  i think we should create this SG in here instead of importing it.

ec2/sc-ec2-linux-jumpcloud-notebook.yaml
```diff
@@ -235,6 +235,8 @@ Resources:
           SetupJumpcloud:
             - install_jc
             - config_jc
+          WriteApacheConf:
+            - WriteApacheConf
```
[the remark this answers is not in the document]
ahayden: I went with `SetupApacheProxy`

ec2/sc-ec2-linux-jumpcloud-notebook.yaml
```diff
@@ -235,6 +247,8 @@ Resources:
           SetupJumpcloud:
             - install_jc
             - config_jc
+          SetupApacheProxy:
```
zaro0508, line 250: You need to add this to `--configsets` here https://github.com/Sage-Bionetworks/service-catalog-library/blob/5ae7de8de55d02b31a5ae938d0736404e09b0f4d/ec2/sc-ec2-linux-jumpcloud-notebook.yaml#L269

ahayden: Sounds good- Should this script be able to handle the auto-reloader cloudformation hook? It looks like all boot scripts are hooked into that event, though this one would fail.

zaro0508: what you've done will not work on a CFN update that replaces an instance.  i think we need to find a solution for this however this PR is also blocking @tthyer 's work so i think we can do a follow on PR for fix.

ahayden: I've updated the utils script to handled rewrites at any time... https://github.com/Sage-Bionetworks/service-catalog-utils/pull/11

That requires re-tagging a utils release, I think

PR #214, opened 2020-07-08
Add target group, alb listener rule-making custom resource, and new NotebookConnectionUri

tthyer:
These are the changes to the notebook template required to make ALB routing work.

Depends on https://github.com/Sage-Bionetworks-IT/cfn-cr-alb-rule/pull/1 https://github.com/Sage-Bionetworks/synapse-login-scipool/pull/40

### DIFF
--- a/ec2/sc-ec2-linux-jumpcloud-notebook.yaml
+++ b/ec2/sc-ec2-linux-jumpcloud-notebook.yaml
@@ -404,7 +404,6 @@ Resources:
   EC2TargetGroup:
     Type: AWS::ElasticLoadBalancingV2::TargetGroup
     Properties:
-      Name: !Sub 'TargetGroup-${LinuxInstance}'
       HealthCheckIntervalSeconds: 30
       HealthCheckProtocol: HTTPS
       HealthCheckTimeoutSeconds: 15
--- a/ec2/sc-ec2-linux-jumpcloud-notebook.yaml
+++ b/ec2/sc-ec2-linux-jumpcloud-notebook.yaml
@@ -456,4 +456,8 @@ Outputs:
     Value: !Sub "https://console.aws.amazon.com/ec2/v2/home?region=${AWS::Region}#Instances:search=${LinuxInstance}"
   NotebookConnectionURI:
     Description: 'Test link for ALB routing'
-    Value: !Sub 'https://connect.scipooldev.org/${LinuxInstance}/'
+    Value: !Sub
+      - '${ALBConnectionURI}/${Path}/'
+      - ALBConnectionURI: !ImportValue
+          'Fn::Sub': '${AWS::Region}-alb-notebook-access-ConnectionURI'
+        Path: !Ref LinuxInstance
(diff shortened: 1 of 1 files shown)


ec2/sc-ec2-linux-jumpcloud-notebook.yaml
```diff
@@ -388,6 +400,42 @@ Resources:
           Value: "yes"
         - Key: "PatchGroup"
           Value: "prod-default"
+
+  EC2TargetGroup:
+    Type: AWS::ElasticLoadBalancingV2::TargetGroup
+    Properties:
+      Name: !Sub 'TargetGroup-${LinuxInstance}'
```
zaro0508, line 407: any reason why you need to explicitly name this?  would be preferabe to let CFN name stuff for us.

ec2/sc-ec2-linux-jumpcloud-notebook.yaml
```diff
@@ -388,6 +400,42 @@ Resources:
...
+      HealthCheckIntervalSeconds: 30
+      HealthCheckProtocol: HTTP
```
ahayden, line 409: This type should actually be HTTPS or it will create SSL connect errors in the web server logs, and probably show up as healthcheck failed.

ec2/sc-ec2-linux-jumpcloud-notebook.yaml
```diff
@@ -404,9 +452,9 @@ Outputs:
...
     Value: !Sub "https://${AWS::Region}.console.aws.amazon.com/systems-manager/session-manager/${LinuxInstance}?region=${AWS::Region}"
-  NotebookConnectionURI:
-    Description: 'Notebook server login page'
-    Value: !Sub "http://${LinuxInstance.PrivateIp}:8787"
   EC2ConsoleURI:
     Description: 'Check your instance status with this link to the AWS Console'
     Value: !Sub "https://console.aws.amazon.com/ec2/v2/home?region=${AWS::Region}#Instances:search=${LinuxInstance}"
+  NotebookConnectionURI:
+    Description: 'Test link for ALB routing'
+    Value: !Sub 'https://connect.scipooldev.org/${LinuxInstance}/'
```
ahayden, line 460: This value actually gets exported by the ALB template if you want to use it: `${AWS::Region}-alb-notebook-access-ConnectionURI`  

As it is, the domain of the dev and prod will be different, so if this template is meant to go to prod (I see `internalpoolvpc`) it will need to use `scipoolprod.org` for notebooks.

ec2/sc-ec2-linux-jumpcloud-notebook.yaml
tthyer: Removed name to avoid replacement of target group when the instance is replaced. Reference: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-elasticloadbalancingv2-targetgroup.html#cfn-elasticloadbalancingv2-targetgroup-name

ec2/sc-ec2-linux-jumpcloud-notebook.yaml
tthyer: Fixed

ec2/sc-ec2-linux-jumpcloud-notebook.yaml
tthyer: Good catch, I switched to the export

PR #218, opened 2020-07-29
Add cfn signals to EC2 products

ahayden:
This change adds a Cloudformation signal counter to common instance templates so that the 'complete' status is reflected in the provisioned product only when all commands in the instance metadata bootstrapping complete successfully. 

This solution works with all current product types.

### DIFF
--- a/ec2/sc-ec2-linux-jumpcloud-notebook.yaml
+++ b/ec2/sc-ec2-linux-jumpcloud-notebook.yaml
@@ -296,9 +296,13 @@ Resources:
               group: root
           commands:
             01_enable_cfn-hup:
-              command: "/bin/systemctl enable cfn-hup.service"
+              command: !Sub |
+                /bin/systemctl enable cfn-hup.service ;\
+                /opt/aws/bin/cfn-signal -e $? --stack ${AWS::StackName} --resource LinuxInstance --region ${AWS::Region}'
             02_start_cfn-hup:
-              command: "/bin/systemctl start cfn-hup.service"
+              command: !Sub |
+                /bin/systemctl start cfn-hup.service ;\
+                /opt/aws/bin/cfn-signal -e $? --stack ${AWS::StackName} --resource LinuxInstance --region ${AWS::Region}'
         set_env_vars:
           files:
             /opt/sage/bin/make_env_vars_file.sh:
@@ -308,7 +312,9 @@ Resources:
               group: "root"
           commands:
             01_make_env_vars_file:
-              command: "/bin/bash /opt/sage/bin/make_env_vars_file.sh"
+              command: !Sub |
+                /bin/bash /opt/sage/bin/make_env_vars_file.sh ;\
+                /opt/aws/bin/cfn-signal -e $? --stack ${AWS::StackName} --resource LinuxInstance --region ${AWS::Region}'
               env:
                 AWS_REGION: !Ref AWS::Region
                 STACK_NAME: !Ref AWS::StackName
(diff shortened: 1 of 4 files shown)


ec2/sc-ec2-linux-jumpcloud-notebook.yaml
```diff
@@ -373,6 +383,10 @@ Resources:
           Value: "yes"
         - Key: "PatchGroup"
           Value: "prod-default"
+    CreationPolicy:
+      ResourceSignal:
+        Count: 7 #Must equal the number of calls to cfn-signal during provisioning
```
zaro0508, line 388: I actually think we only need 1 signal, the one from `UserData:`.  That signal is sent after all init sections have completed.

ec2/sc-ec2-windows-jumpcloud.yaml
```diff
@@ -238,6 +238,9 @@ Resources:
           Value: "yes"
         - Key: "PatchGroup"
           Value: "prod-default"
+    CreationPolicy:
+      ResourceSignal:
+        Timeout: PT10M
```
zaro0508, line 243: i think this may need 20 mins

ec2/sc-ec2-linux-jumpcloud-notebook.yaml
```diff
@@ -373,6 +372,9 @@ Resources:
           Value: "yes"
         - Key: "PatchGroup"
           Value: "prod-default"
+    CreationPolicy:
+      ResourceSignal:
+        Timeout: PT10M
```
zaro0508, line 377: do you think this is enough of a timeout for notebook?

ec2/sc-ec2-linux-jumpcloud-notebook.yaml
```diff
@@ -341,7 +340,7 @@ Resources:
               group: "root"
           commands:
             01_rewrite_apache_conf:
-              command: "/bin/bash /opt/sage/bin/apache_conf_rewrite.sh"
+              command: "/bin/bash /opt/sage/bin/apache_conf_rewrite.sh "
```
zaro0508, line 343: is this change intentional?

ec2/sc-ec2-linux-jumpcloud.yaml
```diff
@@ -295,12 +295,11 @@ Resources:
         TagRootVolume:
           commands:
             01_tag_root_disk:
-              command: !Join
-                - ''
-                - - ". /opt/sage/bin/instance_env_vars.sh;"
-                  - "/usr/bin/aws ec2 create-tags --region $AWS_REGION --resources $ROOT_DISK_ID "
-                  - "--tags Key=Name,Value=$EC2_INSTANCE_ID-root Key=cloudformation:stack-name,Value=$STACK_NAME "
-                  - "Key=Department,Value=$DEPARTMENT Key=Project,Value=$PROJECT Key=OwnerEmail,Value=$OWNER_EMAIL"
+              command: !Sub |
```
zaro0508, line 298: if you refactor this for one file why not refactor for all files?

ahayden: There are only a few stings like this., and I've repeated it across the templates that have this string. If we want to refactor to simply all the templates, we can make a new ticket about it.

ec2/sc-ec2-linux-jumpcloud-notebook.yaml
ahayden: Was not!

ec2/sc-ec2-windows-jumpcloud.yaml
ahayden: This value was fine during testing, but I can set it to 20.

ec2/sc-ec2-linux-jumpcloud-notebook.yaml
ahayden: The duration was fine during testing, but I've increased it to 20 as I don't think it will harm anything.

ec2/sc-ec2-linux-jumpcloud-notebook.yaml
ahayden: Confirmed

PR #221, opened 2020-08-16
Add the ssm-user so that it can be added to the docker group

tthyer:
This isn't what I would ideally like, but it works. 
- creates ssm-user
- adds to docker group
- adds ssm-user as sudoer

I've tested this a bit in scipooldev and it appears to work -- I can connect through the ssm browser session as the ssm-user, can sudo without being prompted for a password, and ssm-user is a member of docker.

I would prefer to be able to configure things about the ssm-user that the ssm agent will create on login, but could find no way to do that.

I am creating the user and setting it up directly in UserData because when I attempt to set it up through a cfn-init config set command, the command appears to succeed, but the user doesn't exist after the command, like something that happens in a subshell.

Reminder, I will be out on Monday -- if we want to fix the bug faster, please merge after approval.

### DIFF
--- a/ec2/sc-ec2-linux-jumpcloud-workflows.yaml
+++ b/ec2/sc-ec2-linux-jumpcloud-workflows.yaml
@@ -219,8 +219,6 @@ Resources:
           SetTags:
             - TagRootVolume
             - TagInstance
-          SetupDocker:
-            - add_docker_user
         cfn_hup_service:
           files:
             /etc/cfn/cfn-hup.conf:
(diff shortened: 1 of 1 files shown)


ec2/sc-ec2-linux-jumpcloud-workflows.yaml
```diff
@@ -317,7 +311,10 @@ Resources:
       UserData:
         Fn::Base64: !Sub |
           #!/bin/bash
-          /opt/aws/bin/cfn-init -v --stack ${AWS::StackName} --resource LinuxInstance --configsets SetupCfn,SetEnv,SetTags,SetupDocker --region ${AWS::Region}
+          /usr/sbin/useradd -m ssm-user -G docker
```
zaro0508, line 314: could we add a comment about why this is setup in UserData instead of `AWS::CloudFormation::Init`?  some yahoo, like me, may try to refactor and move it there in the future.

PR #222, opened 2020-08-19
[SC-140] multiple synapse bucket owners

zaro0508:
With the fix to PLFM-6358 we can now refactor to allow users to
provision Synapse bucket with multiple owners.

### DIFF
--- a/s3/sc-s3-synapse-ra.yaml
+++ b/s3/sc-s3-synapse-ra.yaml
@@ -13,13 +13,17 @@ Metadata:
       - Label:
           default: S3 Bucket Policy Configuration
         Parameters:
-          - SynapseUserName
+          - SynapseIDs
           - S3UserARNs
 Parameters:
-  SynapseUserName:
-    Type: String
-    Description: Synapse user for bucket
+  SynapseIDs:
+    Type: CommaDelimitedList
+    Description: >-
+      (Optional) List of Synapse bucket user or team owners
     Default: ""
+    ConstraintDescription: >-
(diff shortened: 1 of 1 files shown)


s3/sc-s3-synapse-ra.yaml
```diff
@@ -13,13 +13,17 @@ Metadata:
...
+          - SynapseIDs
           - S3UserARNs
 Parameters:
-  SynapseUserName:
-    Type: String
-    Description: Synapse user for bucket
+  SynapseIDs:
+    Type: CommaDelimitedList
+    Description: >-
+      (Optional) List of Synapse bucket user or team owners
```
tthyer, line 22: This can't be optional, as it is required to link to synapse.

brucehoff: Yes, I agree with Tess.

zaro0508: I disagree.  Users should be able to provision a bucket without a synapse ID.  They can just update this field with one or more synapse IDs at a later time.

tthyer: But why would anyone choose to provision a Synapse bucket without linking it when there's also the choice for a non-Synapse bucket? The users don't know how this works behind the scenes. If you let them provision a Synapse bucket without specifying the owner, the linkup with Synapse won't work and they won't know why.

zaro0508: RTFD? 😄

PR #226, opened 2020-09-23
[SC-168] move bash scipt to template

zaro0508:
In an effort to remove a dependency on Sage-Bionetworks/service-catalog-utils
repo and its inherant transitive dependencies.  We move the script into
the cloudformation template.


zaro0508:
This is a bad idea.

PR #227, opened 2020-09-24
[SC-168] Remove apply_name_tag.sh script

zaro0508:
The functionality to apply product name tag and access role approved tag
has beeen moved to the Synapse tagger.  Renoving it from here.

depends on Sage-Bionetworks-IT/cfn-cr-synapse-tagger#13


tthyer (approved):
Nice -- this is going to be so much cleaner. Hopefully avoid writing powershell in future, too.

brucehoff (approved):

ahayden (approved):
`apply_name_tag.sh` theoretically only needs to run when the instance boots the first time, and the script will stay on the instance for ever... I don't think there are any issues with migrating or supporting already-provisioned instances.

PR #229, opened 2020-09-28
[SC-133] Remove VpcName input

zaro0508:
This change removes the VpcName input parameter users see when
provisioning EC2s in the service catalog. We automatically set the
VPC specific parameters based on which account the EC2 is provisioned
in.

### DIFF
--- a/ec2/sc-ec2-linux-jumpcloud-notebook.yaml
+++ b/ec2/sc-ec2-linux-jumpcloud-notebook.yaml
@@ -18,17 +18,20 @@ Metadata:
         default: Disk Size
 
 Mappings:
-
   NotebookTypes:
     Rstudio:
       AMIID: "ami-051310d409a32a5aa"  # https://github.com/Sage-Bionetworks-IT/packer-rstudio/tree/v2.0.0
+  AccountToImportParams:
+    "465877038949":   # org-sagebase-scipooldev
+      PrivateSubnet: "us-east-1-cesspoolvpc-PrivateSubnet"
+      VpnSecurityGroup: "us-east-1-cesspoolvpc-VpnSecurityGroup"
+      VPCId: "us-east-1-cesspoolvpc-VPCId"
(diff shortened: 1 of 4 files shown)


ec2/sc-ec2-linux-jumpcloud-notebook.yaml
```diff
@@ -18,18 +18,21 @@ Metadata:
         default: Disk Size
 
 Mappings:
-
   NotebookTypes:
     Rstudio:
       AMIID: "ami-051310d409a32a5aa"  # https://github.com/Sage-Bionetworks-IT/packer-rstudio/tree/v2.0.0
+  AccountToImportParams:
```
ahayden, line 24: It would be even better to move the map to another template and export it.. then import the key/value every time we need to use it.

ahayden (approved):
Good idea!

ec2/sc-ec2-linux-jumpcloud-notebook.yaml
zaro0508: > It would be even better to move the map to another template and export it.. then import the key/value every time we need to use it.

hmm, interesting idea @ahayden.  i'm not sure what that would look like.  Maybe you can do a PR for it so we can discuss your idea in the PR?

tthyer: I like that idea, and an AWS::Include can be used in the mappings section of the template (just not in the parameters section)

tthyer (approved):
Great, really glad to see this improvement. Remember, we do need to change the documentation when this happens, and let Anna know to change the docs on the portal, too.

ec2/sc-ec2-linux-jumpcloud-notebook.yaml
zaro0508: still not seeing how `AWS::Include` would help make this better.  looks like you guys know something i cannot see.  I'm looking forward to a PR for enlightenment.

PR #232, opened 2020-10-07
[SC-209] add strides account params

zaro0508:
Add strides account to mapping so we can automatically set the
subnets for EC2 instances.


tthyer (approved):
LGTM. I still like @ahayden's suggestion of making `AccountToImportParams` into an include transform, though.

zaro0508:
> LGTM. I still like @ahayden's suggestion of making `AccountToImportParams` into an include transform, though.

Again, I don't know what that looks like.  contributions are always welcome though 😄

xschildw (approved):

PR #242, opened 2021-03-10
[SC-306] lambda docker runner product

zaro0508:
Create a SC product to allow users to provision a lambda that will
run a docker container manually or on a schedule.

depends on https://github.com/Sage-Bionetworks/aws-infra/pull/281 https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/129

### DIFF
--- a/lambda/docker-runner.yaml
+++ b/lambda/docker-runner.yaml
@@ -58,7 +58,7 @@ Parameters:
   ImageUri:
     Type: String
     Description: >
-      The URI of the Amazon ECR) repository for the Lambda function's container image.
+      The URI of the Amazon ECR repository for the Lambda function's container image.
       Example: 111111111.dkr.ecr.us-east-1.amazonaws.com/MY-DOCKER:latest
   EntryPoints:
     Type: CommaDelimitedList
--- a/lambda/docker-runner.yaml
+++ b/lambda/docker-runner.yaml
@@ -156,6 +156,7 @@ Resources:
         EntryPoint: !If [HasEntryPoints, !Ref EntryPoints, !Ref 'AWS::NoValue']
         WorkingDirectory: !If [HasWorkingDirectory, !Ref WorkingDirectory, !Ref 'AWS::NoValue']
       MemorySize: !Ref MemorySize
+      KmsKeyArn: !ImportValue {'Fn::Sub': '${AWS::Region}-docker-runner-kms-key-KmsKeyArn'}
       Environment:
         Variables:
           EFS_PATH: !Ref EfsPath
--- a/lambda/docker-runner.yaml
+++ b/lambda/docker-runner.yaml
@@ -1,9 +1,7 @@
-# This template depends on having an AWS EFS in place which can be deployed with
-# https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/EFS/efs.yaml
+# Based off of https://github.com/aws-samples/aws-lambda-efs-samples
 AWSTemplateFormatVersion: '2010-09-09'
-Transform: S3Objects
-Description: >-
-  Docker runner product
+Transform: AWS::Serverless-2016-10-31
+Description: Lambda to run a docker container
 Metadata:
   AWS::CloudFormation::Interface:
     ParameterGroups:
@@ -11,42 +9,65 @@ Metadata:
           default: Container Image
         Parameters:
           - ImageUri
-          - EntryPoint
+          - EntryPoints
           - Command
           - WorkingDirectory
           - MemorySize
       - Label:
           default: Synapse
         Parameters:
-          - SynapseUser
-          - SynapsePassword
+          - SynapsePAT
       - Label:
           default: Execution Trigger
         Parameters:
           - EnableSchedule
           - Schedule
+      - Label:
+          default: EFS
+        Parameters:
+          - PerformanceMode
+          - Encrypted
+          - EfsPath
 Parameters:
+  PerformanceMode:
+    Type: String
+    Default: "generalPurpose"
+    Description: The EFS disk performance mode
+    AllowedValues:
+      - generalPurpose
+      - maxIO
+  Encrypted:
+    Type: String
+    Default: true
+    Description: True to enable EFS encryption, false to disable.
+    AllowedValues:
+      - true
+      - false
+  EfsPath:
+    Type: String
+    Description: Path to access EFS file store.
+    Default: /mnt/efs
   MemorySize:
     Type: Number
     MinValue: 128
     MaxValue: 10240
     Default: 128
     Description: >
-      The amount of memory available to the function at runtime. Increasing the function's
-      memory also increases its CPU allocation.
+      The amount of memory (in MB) available to the function at runtime.
+      Increasing the function's memory also increases its CPU allocation.
   ImageUri:
     Type: String
     Description: >
       The URI of the Amazon ECR) repository for the Lambda function's container image.
       Example: 111111111.dkr.ecr.us-east-1.amazonaws.com/MY-DOCKER:latest
-  EntryPoint:
-    Type: String
+  EntryPoints:
+    Type: CommaDelimitedList
     Default: ""
-    Description: (Optional) The entry point to the application in the docker container
+    Description: (Optional) Comma separated list of entry points to docker container
   Command:
     Type: CommaDelimitedList
     Default: ""
-    Description: (Optional) Command separated list of parameters to pass in with docker ENTRYPOINT
+    Description: (Optional) Comma separated list of parameters to pass in with docker ENTRYPOINT
   WorkingDirectory:
     Type: String
     Default: ""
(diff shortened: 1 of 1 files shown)


lambda/docker-runner.yaml
```diff
@@ -0,0 +1,111 @@
...
+    MinValue: 128
+    MaxValue: 10240
+    Default: 128
+    Description: >
+      The amount of memory available to the function at runtime. Increasing the function's
+      memory also increases its CPU allocation.
+  ImageUri:
+    Type: String
+    Description: >
+      The URI of the Amazon ECR) repository for the Lambda function's container image.
```
ahayden, line 40: I believe we want to pull images from the Synapse reg... is this really limited to ECR?

lambda/docker-runner.yaml
```diff
@@ -0,0 +1,111 @@
...
+          - ImageUri
+          - EntryPoint
+          - Command
+          - WorkingDirectory
+          - MemorySize
+      - Label:
+          default: Synapse
+        Parameters:
+          - SynapseUser
+          - SynapsePassword
```
ahayden, line 22: We should not call this a password if we want people to use other types of secrets... which I think is the case.

lambda/docker-runner.yaml
```diff
@@ -0,0 +1,111 @@
...
+          'Fn::Sub': '${AWS::Region}-docker-runner-efs-AccessPointArn'
+        MemorySize: !Ref MemorySize
+        ImageUri: !Ref ImageUri
+        Command: !If [HasCommand, !Ref Command, !Ref 'AWS::NoValue']
+        EntryPoint: !If [HasEntryPoint, !Ref EntryPoint, !Ref 'AWS::NoValue']
+        WorkingDirectory: !If [HasWorkingDirectory, !Ref WorkingDirectory, !Ref 'AWS::NoValue']
+        EnableSchedule: !Ref EnableSchedule
+        Schedule: !Ref Schedule
+        SynapseUser: !Ref SynapseUser
+        SynapsePassword: !Ref SynapsePassword
```
ahayden, line 107: Do we really want to export a secret like this? Seems excessively permissive

ahayden (changes requested):
There must a be a better way to handle the Synapse secrets...

lambda/docker-runner.yaml
```diff
@@ -0,0 +1,111 @@
+# This template depends on having an AWS EFS in place which can be deployed with
+# https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/EFS/efs.yaml
```
brucehoff, line 2: Since EFS can be expensive we need to make sure that it's tagged with the user's ID

lambda/docker-runner.yaml
```diff
@@ -0,0 +1,111 @@
...
+Transform: S3Objects
+Description: >-
+  Docker runner product
+Metadata:
+  AWS::CloudFormation::Interface:
+    ParameterGroups:
+      - Label:
+          default: Container Image
+        Parameters:
+          - ImageUri
```
brucehoff, line 13: A minor point, but a reference to a Docker image is not a URI.  I suggest calling this parameter "ImageReference".  It should allow a reference to a Docker repository image or a commit and takes the form "NAME [:TAG|@DIGEST]".  See https://docs.docker.com/engine/reference/commandline/pull/.  Examples of valid references include:
ubuntu
rocker/r-base
sagebionetworks/some/repo
sagebionetworks/some/repo:1.1
docker.synapse.org/syn123456/my/repo
docker.synapse.org/syn123456/my/repo@sha256:b986e16563abd60ffda0f1ded1e45404ec9296d4b3c7f31bc4f938879f4627d4

lambda/docker-runner.yaml
```diff
@@ -0,0 +1,111 @@
...
+          default: Execution Trigger
+        Parameters:
+          - EnableSchedule
+          - Schedule
+Parameters:
+  MemorySize:
+    Type: Number
+    MinValue: 128
+    MaxValue: 10240
+    Default: 128
```
ahayden, line 33: This default is probably too low... but arriving at the right default is probably trial and error

lambda/docker-runner.yaml
```diff
@@ -0,0 +1,111 @@
...
+    Default: false
+    AllowedValues:
+      - true
+      - false
+  Schedule:
+    Description: >
+      Schedule to execute the docker, can be a rate or a cron schedule. Format at
+      https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html
+    Type: String
+    Default: cron(0 09 ? * MON *)  # Run every Monday at 9am
```
ahayden, line 67: This default should probably be run once or we may get a bunch of jobs running on Monday at 9 am repeatedly on accident

lambda/docker-runner.yaml
```diff
@@ -0,0 +1,111 @@
...
+          'Fn::Sub': '${AWS::Region}-docker-runner-efs-EfsPath'
+        AccessPointArn: !ImportValue
+          'Fn::Sub': '${AWS::Region}-docker-runner-efs-AccessPointArn'
+        MemorySize: !Ref MemorySize
+        ImageUri: !Ref ImageUri
+        Command: !If [HasCommand, !Ref Command, !Ref 'AWS::NoValue']
+        EntryPoint: !If [HasEntryPoint, !Ref EntryPoint, !Ref 'AWS::NoValue']
+        WorkingDirectory: !If [HasWorkingDirectory, !Ref WorkingDirectory, !Ref 'AWS::NoValue']
+        EnableSchedule: !Ref EnableSchedule
+        Schedule: !Ref Schedule
```
ahayden, line 105: It would be better to make this a NoValue if not defined, rather than a default schedule that repeats.

lambda/docker-runner.yaml
```diff
@@ -0,0 +1,111 @@
...
+    Description: >
+      The amount of memory available to the function at runtime. Increasing the function's
+      memory also increases its CPU allocation.
+  ImageUri:
+    Type: String
+    Description: >
+      The URI of the Amazon ECR) repository for the Lambda function's container image.
+      Example: 111111111.dkr.ecr.us-east-1.amazonaws.com/MY-DOCKER:latest
+  EntryPoint:
+    Type: String
```
brucehoff, line 43: I'm pretty sure (but not 100% sure) that EntryPoint should be a list, like CMD is, https://phoenixnap.com/kb/docker-run-override-entrypoint

lambda/docker-runner.yaml
```diff
@@ -0,0 +1,111 @@
...
+      - true
+      - false
+  Schedule:
+    Description: >
+      Schedule to execute the docker, can be a rate or a cron schedule. Format at
+      https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html
+    Type: String
+    Default: cron(0 09 ? * MON *)  # Run every Monday at 9am
+    ConstraintDescription: "Use schedule format: https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html"
+  SynapseUser:
```
brucehoff, line 69: Again, pls replace user+pw with access token

brucehoff (changes requested):

lambda/docker-runner.yaml
```diff
@@ -0,0 +1,111 @@
...
+    Type: String
+    Description: >
+      The URI of the Amazon ECR) repository for the Lambda function's container image.
+      Example: 111111111.dkr.ecr.us-east-1.amazonaws.com/MY-DOCKER:latest
+  EntryPoint:
+    Type: String
+    Default: ""
+    Description: (Optional) The entry point to the application in the docker container
+  Command:
+    Type: CommaDelimitedList
```
zimingd, line 47: Is this the list of arguments passed to the entrypoint executable?

lambda/docker-runner.yaml
```diff
@@ -0,0 +1,111 @@
...
+        Parameters:
+          - SynapseUser
+          - SynapsePassword
+      - Label:
+          default: Execution Trigger
+        Parameters:
+          - EnableSchedule
+          - Schedule
+Parameters:
+  MemorySize:
```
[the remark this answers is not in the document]
zaro0508: Again, just taken from AWS terminology. https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html#sam-function-memorysize

How about we add a description that will appear in the UI?  Something like this.. `Memory (in MB) allocated for the lambda.`

lambda/docker-runner.yaml
```diff
@@ -0,0 +1,111 @@
...
+        Parameters:
+          - ImageUri
+          - EntryPoint
+          - Command
+          - WorkingDirectory
+          - MemorySize
+      - Label:
+          default: Synapse
+        Parameters:
+          - SynapseUser
```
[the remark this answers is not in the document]
zaro0508: ahh yes, I forget we want to let user pass in the PAT.  will fix.

lambda/docker-runner.yaml
zaro0508: will replace with a PAT

lambda/docker-runner.yaml
zaro0508: please make a suggestion

lambda/docker-runner.yaml
zaro0508: hmm, from reading the docs again i think ImageUri can only pull from ECR.   Making a note to myself to  test getting an image from public dockerhub respository.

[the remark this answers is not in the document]
zaro0508: I'm not sure what type of repo the Synapse one runs in but I think ImageUri only supports pull from EC2 public and private repos.  for ECR private repos we can just use a AWS service account to access the repo.

lambda/docker-runner.yaml
zaro0508: it's the `CMD` that sets the default parameters.

lambda/docker-runner.yaml
zaro0508: The schedule is only enabled if the `EnableSchedule` parameter is true.   Please suggest something if you don't like the default.

lambda/docker-runner.yaml
zaro0508: will do.

lambda/docker-runner.yaml
zaro0508: The schedule is not enabled by default it will only run if `EnableSchedule` is true.

lambda/docker-runner.yaml
zaro0508: I have tried pointing this to a test image on my public dockerhub repository (https://hub.docker.com/r/zaro0508/docker-lambda) but I cannot make it work.  Also I've checked the AWS docs again and I cannot find any references to lambda support for pulling images from dockerhub :(

brucehoff: Thinking about it, it makes sense that AWS Lambda would not want to be throttled by DockerHub.  Requiring you to use an AWS registry means they don't get dragged into the pay-or-be-throttled morass.

Please check with Allaway and other key users to see if this limitation is prohibitive.  Maybe they are fine with putting their images into an AWS registry.

lambda/docker-runner.yaml
ahayden: I suppose if the `EnableScheduled` parameter is false by default, it's not an issue. This is in the local TZ of the region, right?

zaro0508: lambda is on UTC time. https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html
```
TZ – The environment's time zone (UTC). The execution environment uses NTP to synchronize the system clock.
```

lambda/docker-runner.yaml
```diff
@@ -0,0 +1,111 @@
...
+  AWS::CloudFormation::Interface:
+    ParameterGroups:
+      - Label:
+          default: Container Image
+        Parameters:
+          - ImageUri
+          - EntryPoint
+          - Command
+          - WorkingDirectory
+          - MemorySize
```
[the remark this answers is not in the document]
zaro0508: added `in MB` in this parameter's description below.

lambda/docker-runner.yaml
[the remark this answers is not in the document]
zaro0508: added `in MB` in the description for this parameter

lambda/docker-runner.yaml
zaro0508: replaced with a synapse personal access token `SynapsePAT`

lambda/docker-runner.yaml
zaro0508: EFS is no longer a shared resource.  We now create an EFS for each lambda that is provisioned.  Each EFS will be contain Department and Projects tags.

lambda/docker-runner.yaml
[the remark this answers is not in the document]
zaro0508: replaced with `SynapsePAT` parameter

lambda/docker-runner.yaml
zaro0508: changed to a list

lambda/docker-runner.yaml
zaro0508: done

lambda/docker-runner.yaml
brucehoff: An EC2 t3.micro uses 1 GB so why not make this the same (1024 MB)?

brucehoff (changes requested):

lambda/docker-runner.yaml
zaro0508: done

brucehoff (approved):

lambda/docker-runner.yaml
```diff
@@ -0,0 +1,202 @@
...
+      PerformanceMode: !Ref PerformanceMode
+      Encrypted: !Ref Encrypted
+      FileSystemPolicy:
+        Version: "2012-10-17"
+        Statement:
+          - Effect: "Allow"
+            Action:
+              - "elasticfilesystem:ClientMount"
+            Principal:
+              AWS: "*"
```
ahayden, line 121: This allows any account principal to mount the EFS AccessPoint created for this lambda? I think there needs to be a condition here to limit EFS access to the intended lambda, or use filesystem permissions (ie. uid/gid) on EFS to manage access to the files stored on it.

lambda/docker-runner.yaml
```diff
@@ -0,0 +1,202 @@
...
+      - true
+      - false
+  Schedule:
+    Description: >
+      Schedule to execute the docker, can be a rate or a cron schedule. Format at
+      https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html
+    Type: String
+    Default: cron(0 09 ? * MON *)  # Run every Monday at 9am
+    ConstraintDescription: "Use schedule format: https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html"
+  SynapsePAT:
```
ahayden, line 90: This will store a Synapse secret in plaintext as a parameter to a cfn stack. The parameters can be read by the admin role in any scipool account, but not by communitymanagers or endusers (they can see the plaintext of their own stack parameters if they know the full stack id). I think this is still excessive, since a PAT is long lived and should be readable only to the Synapse account holder. `secretsmanager` and limits to IAM roles are probably the way to do it properly...

lambda/docker-runner.yaml
```diff
@@ -0,0 +1,202 @@
...
+      ImageUri: !Ref ImageUri
+      ImageConfig:
+        Command: !If [HasCommand, !Ref Command, !Ref 'AWS::NoValue']
+        EntryPoint: !If [HasEntryPoints, !Ref EntryPoints, !Ref 'AWS::NoValue']
+        WorkingDirectory: !If [HasWorkingDirectory, !Ref WorkingDirectory, !Ref 'AWS::NoValue']
+      MemorySize: !Ref MemorySize
+      Environment:
+        Variables:
+          EFS_PATH: !Ref EfsPath
+          SYNAPSE_PAT: !If [HasSynapsePAT, !Ref SynapsePAT, !Ref 'AWS::NoValue']
```
ahayden, line 170: The Synapse secret (each input env variable) is encrypted with KMS using the default key for this account, sent to wherever the lambda runs, and decrypted by the lambda runner when invoked.

ahayden (changes requested):
I think storing the Synapse PAT as a parameter is a problem... We shouldn't let people start using it until the PAT can't be read by anyone but the Synapse account holder.

lambda/docker-runner.yaml
```diff
@@ -0,0 +1,202 @@
...
+    Type: String
+    Default: true
+    Description: True to enable EFS encryption, false to disable.
+    AllowedValues:
+      - true
+      - false
+  EfsPath:
+    Type: String
+    Description: Path to access EFS file store.
+    Default: /mnt/efs
```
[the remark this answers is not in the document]
brucehoff: Note that there is another level of abstraction/indirection:  The `docker run` command includes a volume mount, e.g.:
```
docker run -v /mnt/efs:/path/inside/container ...
```
so the container image need not reference the host mount point.  So portability may not be an issue

lambda/docker-runner.yaml
```diff
@@ -0,0 +1,202 @@
...
+    AllowedValues:
+      - generalPurpose
+      - maxIO
+  Encrypted:
+    Type: String
+    Default: true
+    Description: True to enable EFS encryption, false to disable.
+    AllowedValues:
+      - true
+      - false
```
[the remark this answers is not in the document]
zaro0508: probably not but it's not something that users control.  The SC service will always set this to true.

lambda/docker-runner.yaml
zaro0508: there is no KMS encryption going on.

lambda/docker-runner.yaml
zaro0508: removed.  we don't need this since access to mount  filesystem is already provided to the lambda function in EfsPolicy

lambda/docker-runner.yaml
zaro0508: secrets will appear as `****` in CFN parameters and secrets are not outputted so it will never appear in CFN outputs section.

lambda/docker-runner.yaml
ahayden: AWS uses KMS when deploying the lamda

lambda/docker-runner.yaml
zaro0508: by adding `KmsKeyArn` to this template and defining the kms key in PR https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/129  the env vars are no longer visible to any logged in user... https://pasteboard.co/JW6CNXg.png

brucehoff: What do you think, @ahayden ?

zaro0508:
We are replacing this implementation with PR 264

PR #243, opened 2021-03-19
[SC-312] switches ownership of uploaded objects to be that of the bucket

zaro0508:
restrict Put-Object calls to include bucket-owner-full-control
users using the AWS CLI must upload with `--acl bucket-owner-full-control`
to cause the ownership to change

depends on https://github.com/Sage-Bionetworks-IT/cfn-cr-sc-bucket-policy/pull/4

### DIFF
--- a/s3/sc-s3-synapse-ra.yaml
+++ b/s3/sc-s3-synapse-ra.yaml
@@ -65,6 +65,9 @@ Resources:
             AllowedOrigins: ['*']
             AllowedMethods: [GET, POST, PUT, HEAD]
             MaxAge: 3000
+      OwnershipControls:
+        Rules:
+          - ObjectOwnership: BucketOwnerPreferred
   S3BucketPolicy:
     Type: Custom::SCS3BucketPolicy
     Properties:


xschildw (approved):

brucehoff (changes requested):
The change in the PR is not what the title and description of the PR indicate:   The change in the PR switches ownership of uploaded objects to be that of the bucket.  The PR title indicates that the change forces the user to add the 'bucket-owner-full-control' ACL to the object (which does not change ownership but which IS a requirement for the BucketOwnershipPreferred setting to cause the ownership to change.)

brucehoff (approved):
This template doesn't include the bucket policy but I assume the policy includes the requirement that uploaded files give the bucket owner full access control, since that's the other necessary piece.

zaro0508:
> This template doesn't include the bucket policy but I assume the policy includes the requirement that uploaded files give the bucket owner full access control, since that's the other necessary piece.

It actually does, it's in the custom resource (Custom::SCS3BucketPolicy).  We manage the bucket policy with that custom resource and the update for it is in the dependent PR https://github.com/Sage-Bionetworks-IT/cfn-cr-sc-bucket-policy/pull/4

BrunoGrandePhD (approved):

ahayden (approved):

PR #251, opened 2021-07-12
Increase disk size limit for SC workflows product

BrunoGrandePhD:
Depending on the dataset size, 2 TB is limiting given the number of large intermediate files that get generated. I propose increasing the limit to 5 TB (just for the workflows product) to give more breathing room. I need this in case NF Tower isn't set up in time to meet deadlines for processing data in the imCORE project. 

I believe once this PR is merged, a new version of this repository will need to be minted and this [file](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/sceptre/scipool/config/prod/sc-product-ec2-linux-jumpcloud-workflows.yaml#L18) will need to be changed.

### DIFF
--- a/ec2/sc-ec2-linux-docker.yaml
+++ b/ec2/sc-ec2-linux-docker.yaml
@@ -192,7 +192,7 @@ Parameters:
     Type: Number
     Default: 16
     MinValue: 16
-    MaxValue: 2000
+    MaxValue: 5000
 Resources:
   InstanceRole:
     Type: AWS::IAM::Role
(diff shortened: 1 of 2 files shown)


BrunoGrandePhD:
Requesting review from @zaro0508 and @tthyer.

tthyer (approved):
This covers what we already discussed, and I agree 5TB is a good jump to make.

zaro0508 (approved):
only workflow?  don't want to increase in other products?  

Just to let you know we might deprecate this product soon to promote the EC2 docker product.

tthyer:
@BrunoGrandePhD IOW consider also increasing the volume size for sc-ec2-linux-docker.yaml, too

BrunoGrandePhD:
@tthyer @zaro0508: I've increased the limit on the Docker product too!

tthyer (approved):

zaro0508 (approved):

PR #256, opened 2021-08-19
SC-318 updating R-Studio product to use new AMI

brucehoff:
Updating R-Studio product (v 2.1.2 of packer-rstudio) to use new AMI.   As per [SC-318](https://sagebionetworks.jira.com/browse/SC-318), this changes `/etc/security/limits.conf` to allow more concurrent open files.

### DIFF
--- a/ec2/sc-ec2-linux-jumpcloud-notebook.yaml
+++ b/ec2/sc-ec2-linux-jumpcloud-notebook.yaml
@@ -24,7 +24,7 @@ Metadata:
 Mappings:
   NotebookTypes:
     Rstudio:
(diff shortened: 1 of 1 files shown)


zaro0508 (changes requested):
Could you please add some more info to the PR description describing why we are updating the AMI and what the update provides?

zaro0508 (approved):
looks good. to propagate this to the SC you'll also need to create a new service-catalog-library  tag then update that version in https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/sceptre/scipool/config/develop/sc-product-ec2-linux-jumpcloud-notebook.yaml

This is the list of changes for updating the AMI, please verify that it all works.. 
https://github.com/Sage-Bionetworks-IT/packer-rstudio/compare/v2.1.1...v2.1.2

PR #258, opened 2021-09-17
make sure bucket policy depends on bucket

zaro0508:
A bucket delete may fail due to having existing data in the bucket.
When a failure occurs the bucket will remain however the  policy for it
may get removed.  This will cause a state where users will no longer
have access to the bucket.

This change makes sure that the policy deletion only happens after a
bucket has been delete.

### DIFF
--- a/.cfnlintrc
+++ b/.cfnlintrc
@@ -1,4 +1,5 @@
 ignore_checks:
 - W1020
 - W3005
+- W3011
 - E1022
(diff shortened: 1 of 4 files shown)


brucehoff (commented):
I see that the change is to make the Policy 'depend on' the Bucket.  I take it to mean that when the resource is created, first the Bucket is created and afterwards the Policy.  Is that correct? If so, isn't the tear down in the reverse order, i.e., first the Policy is removed and afterwards the Bucket?   That would contradict your comment, "This change makes sure that the policy deletion only happens after a bucket has been delete[d]."

zaro0508:
yes, you are correct @brucehoff so it's back to the drawing board.

PR #261, opened 2021-10-04
[SC-363]  Enable S3 intelligent tierring on buckets

zaro0508:
We SC users provision buckets they use it for a variety of different
things.  We don't really know their usage patters ahead of time.  They
can upload files and leave them around for a long time which can be
expensive.  Enabling intelligent tiering on their buckets may help us
reduce S3 storage costs.  Well, this is what @marcomarasca says anyways.

### DIFF
--- a/s3/sc-s3-encrypted-ra.yaml
+++ b/s3/sc-s3-encrypted-ra.yaml
@@ -51,6 +51,14 @@ Resources:
         ServerSideEncryptionConfiguration:
           - ServerSideEncryptionByDefault:
               SSEAlgorithm: AES256
+      LifecycleConfiguration:
+        Rules:
+          - Id: IntelligentTieringRule
+            Prefix: IntTier
+            Status: Enabled
+            Transitions:
+              - TransitionInDays: 1
+                StorageClass: INTELLIGENT_TIERING
   S3BucketTagger:
     DependsOn: S3BucketPolicy
     Type: Custom::SynapseTagger
--- a/s3/sc-s3-encrypted-ra.yaml
+++ b/s3/sc-s3-encrypted-ra.yaml
@@ -53,8 +53,7 @@ Resources:
               SSEAlgorithm: AES256
       LifecycleConfiguration:
         Rules:
-          - Id: IntelligentTieringRule
-            Prefix: IntTier
+          - Id: IntelligentTieringClassTransitionRule
             Status: Enabled
             Transitions:
               - TransitionInDays: 1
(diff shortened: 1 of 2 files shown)


s3/sc-s3-encrypted-ra.yaml
```diff
@@ -51,6 +51,14 @@ Resources:
         ServerSideEncryptionConfiguration:
           - ServerSideEncryptionByDefault:
               SSEAlgorithm: AES256
+      LifecycleConfiguration:
+        Rules:
+          - Id: IntelligentTieringRule
+            Prefix: IntTier
```
marcomarasca, line 57: The Prefix is used to identifies which objects the rule applies to, in particular it would be an object key prefix so I do not think we need this.

s3/sc-s3-synapse-ra.yaml
```diff
@@ -70,6 +70,14 @@ Resources:
       OwnershipControls:
         Rules:
           - ObjectOwnership: BucketOwnerPreferred
+      LifecycleConfiguration:
+        Rules:
+          - Id: IntelligentTieringRule
+            Prefix: IntTier
```
marcomarasca, line 76: The Prefix is used to identifies which objects the rule applies to, in particular it would be an object key prefix so I do not think we need this.

s3/sc-s3-encrypted-ra.yaml
```diff
@@ -51,6 +51,14 @@ Resources:
         ServerSideEncryptionConfiguration:
           - ServerSideEncryptionByDefault:
               SSEAlgorithm: AES256
+      LifecycleConfiguration:
+        Rules:
+          - Id: IntelligentTieringRule
```
marcomarasca, line 56: I would call this IntelligentTieringClassTransitionRule

s3/sc-s3-synapse-ra.yaml
```diff
@@ -70,6 +70,14 @@ Resources:
       OwnershipControls:
         Rules:
           - ObjectOwnership: BucketOwnerPreferred
+      LifecycleConfiguration:
+        Rules:
+          - Id: IntelligentTieringRule
```
marcomarasca, line 75: I would call this IntelligentTieringClassTransitionRule

marcomarasca (changes requested):

s3/sc-s3-encrypted-ra.yaml
zaro0508: removed

s3/sc-s3-encrypted-ra.yaml
zaro0508: done

s3/sc-s3-synapse-ra.yaml
zaro0508: done

marcomarasca (approved):

PR #262, opened 2021-10-21
SC-45 remove arbitrary Role ARN parameter

brucehoff:
We remove the IAM Role parameter from our two S3 bucket products to remove the security risk of a user giving too permissive access to the bucket.  Users should get access the same way they obtain access to other SC products (like EC2 instances), by obtaining an STS token returned by our system.  This is already possible and the user documentation has already been updated to instruct them to get access this way.

See https://sagebionetworks.jira.com/browse/SC-45 for further discussion.

### DIFF
--- a/s3/sc-s3-encrypted-ra.yaml
+++ b/s3/sc-s3-encrypted-ra.yaml
@@ -11,17 +11,8 @@ Metadata:
       - Label:
           default: S3 Bucket Policy Configuration
         Parameters:
-          - S3UserARNs
           - RestrictSameRegionDownload
 Parameters:
-  S3UserARNs:
-    Type: CommaDelimitedList
-    Description: >-
-      (Optional) User ARNs allowed to access S3 Bucket,
-      in addition to provisioning user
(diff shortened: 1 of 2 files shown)


brucehoff:
> > The `owner.txt` file does not control access by the S3 client, so adding other users to that file does not do what adding IAM ARNs does currently.
> 
> Yes, i understand this. I'm referring to sharing the data in the bucket with multiple users via the synapse web UI or synapse client, not the AWS S3 client.

I see.  Yes, it's a good idea for us to emphasize in our doc's that the first choice (recommended way) for sharing a bucket is to link the bucket to Synapse and then use Synapse's rich/mature feature set to share access with others.  My comments in SC-45 were meant specifically to address the case in which two people want to access a bucket using the S3 client.

> 
> > But I think you are asking, "Could the bucket have a parameter for a list of Synapse users who would be granted S3 access." Is that what you are asking?
> 
> Yes, I'm asking this as well.

That's a good idea, could be done as a separate issue.

PR #264, opened 2021-11-14
[SC-306] Scheduled Jobs product

zaro0508:
Add a SC product to run cron jobs on a schedule.  The cron jobs
will be run as AWS batch jobs.  Users can set the job to run on
a schedule or they can manually trigger the job with the submit
job API.

depends on https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/317 https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/320

### DIFF
--- a/batch/sc-batch-fargate.yaml
+++ b/batch/sc-batch-fargate.yaml
@@ -1,0 +1,190 @@
+AWSTemplateFormatVersion: "2010-09-09"
+Description: 'AWS Batch with Fargate'
+Metadata:
+  AWS::CloudFormation::Interface:
+    ParameterGroups:
+      - Label:
+          default: Container Image
+        Parameters:
+          - Image
+          - Command
+      - Label:
+          default: Synapse
+        Parameters:
+          - SynapsePAT
+      - Label:
+          default: Schedule Job
+        Parameters:
+          - EnableSchedule
+          - Schedule
+Parameters:
+  Image:
+    Type: String
+    Description: >
+      The container image to run, can be from dockerhub or from AWS ECR repository.
+    Default: "debian:latest"
+    ConstraintDescription: >
+      Must be a valid dockerhub or AWS ECR container image
+      Example: debian:latest or 111111111.dkr.ecr.us-east-1.amazonaws.com/MY-IMAGE:latest
+  Command:
+    Type: List<String>
+    Description: >
+      The list of commands to pass to the container image
+    ConstraintDescription: >
+      Must be a list of commands (i.e. ["echo", "hello", "world"])
+  EnableSchedule:
+    Description: >
+      true to run on a schedule, false to disable. If enabled a valid Schedule must be provided
+    Type: String
+    Default: false
+    AllowedValues:
+      - true
+      - false
+  Schedule:
+    Description: >
+      Schedule to execute the docker, can be a rate or a cron schedule. Format at
+      https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html
+    Type: String
+    Default: cron(0 09 ? * MON *)  # Run every Monday at 9am
+    ConstraintDescription: "Use schedule format: https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html"
+  SynapsePAT:
+    Type: String
+    NoEcho: true
+    Default: ""
+    Description: >
+      (Optional) The Synapse personal access token (PAT).
+      https://docs.synapse.org/rest/POST/personalAccessToken.html
+Resources:
+  ComputeEnvironment:
+    Type: AWS::Batch::ComputeEnvironment
+    Properties:
+      Type: MANAGED
+      State: ENABLED
+      ServiceRole: !Ref ServiceRole
+      ComputeResources:
+        Type: FARGATE
+        MaxvCpus: 1
+        Subnets:
+          - !ImportValue us-east-1-scriptrunnervpc-PrivateSubnet
+          - !ImportValue us-east-1-scriptrunnervpc-PrivateSubnet1
+          - !ImportValue us-east-1-scriptrunnervpc-PrivateSubnet2
+        SecurityGroupIds:
+          - !Ref SecurityGroup
+  ServiceRole:
+    Type: AWS::IAM::Role
+    Properties:
+      AssumeRolePolicyDocument:
+        Version: 2012-10-17
+        Statement:
+          - Effect: 'Allow'
+            Principal:
+              Service:
+                - 'batch.amazonaws.com'
+            Action:
+              - 'sts:AssumeRole'
+      ManagedPolicyArns:
+        - 'arn:aws:iam::aws:policy/service-role/AWSBatchServiceRole'
+  JobQueue:
+    Type: AWS::Batch::JobQueue
+    Properties:
+      ComputeEnvironmentOrder:
+        - ComputeEnvironment: !Ref ComputeEnvironment
+          Order: 1
+      Priority: 1
+      State: ENABLED
+  SchedulingPolicy:
+    Type: AWS::Batch::SchedulingPolicy
+    Properties:
+      FairsharePolicy:
+        ShareDecaySeconds: 5
+        ComputeReservation: 2
+  SecurityGroup:
+    Type: AWS::EC2::SecurityGroup
+    Properties:
+      VpcId: !ImportValue us-east-1-scriptrunnervpc-VPCId
+      GroupDescription: !Sub 'SC for ${AWS::StackName}}'
+  JobDefinition:
+    Type: AWS::Batch::JobDefinition
+    Properties:
+      Type: container
+      JobDefinitionName: { Ref: "AWS::StackName" }
+      PropagateTags: True
+      PlatformCapabilities:
+        - FARGATE
+      Timeout:
+        AttemptDurationSeconds: 600
+      RetryStrategy:
+        Attempts: 1
+      ContainerProperties:
+        Command: !Ref Command
+        Image: !Ref Image
+        Environment:
+          - Name: SynapsePAT
+            Value: !Ref SynapsePAT
+        NetworkConfiguration:
+          AssignPublicIp: ENABLED
+        ResourceRequirements:
+          - Type: VCPU
+            Value: 0.5
+          - Type: MEMORY
+            Value: 1024
+        JobRoleArn: !GetAtt JobRole.Arn
+        ExecutionRoleArn: !GetAtt ExecutionRole.Arn
+        LogConfiguration:
+          LogDriver: awslogs
+          Options:
+            "awslogs-group": !Ref LogGroup
+  ExecutionRole:
+    Type: AWS::IAM::Role
+    Properties:
+     AssumeRolePolicyDocument:
+        Version: 2012-10-17
+        Statement:
+          - Effect: 'Allow'
+            Principal:
+              Service:
+                - 'ecs-tasks.amazonaws.com'
+            Action:
+              - 'sts:AssumeRole'
+     ManagedPolicyArns:
+       - 'arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy'
+  JobRole:
+    Type: 'AWS::IAM::Role'
+    Properties:
+      AssumeRolePolicyDocument:
+        Version: 2012-10-17
+        Statement:
+          - Effect: 'Allow'
+            Principal:
+              Service:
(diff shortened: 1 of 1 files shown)


thomasyu888:
Can the job access AWS secrets? What if the image needs environmental variables like synapse PAT?

brucehoff (changes requested):

batch/sc-batch-fargate.yaml
```diff
@@ -0,0 +1,200 @@
...
+Parameters:
+  Image:
+    Type: String
+    Description: >
+      The container image to run, can be from dockerhub or from AWS ECR repository.
+    Default: "debian:latest"
+    ConstraintDescription: >
+      Must be a valid dockerhub or AWS ECR container image
+      Example: debian:latest or 111111111.dkr.ecr.us-east-1.amazonaws.com/MY-IMAGE:latest
+  Command:
```
brucehoff, line 29: In addition to providing a `Command` parameter, providing a list of environment variables would be useful.  If we don't do this I fear there will be a lot of Docker repo's that can't be used with this product.

batch/sc-batch-fargate.yaml
```diff
@@ -0,0 +1,200 @@
+AWSTemplateFormatVersion: "2010-09-09"
+Description: 'AWS Batch with Fargate'
```
thomasyu888, line 2: Is the reason to use fargate because it is a serverless option?

batch/sc-batch-fargate.yaml
```diff
@@ -0,0 +1,200 @@
...
+        - FARGATE
+      Timeout:
+        AttemptDurationSeconds: 600
+      RetryStrategy:
+        Attempts: 1
+      ContainerProperties:
+        Command: !Ref Command
+        Image: !Ref Image
+        Environment:
+          - Name: SynapsePAT
```
thomasyu888, line 132: This key should be `SYNAPSE_AUTH_TOKEN`.  The clients auto recognize this environmental variable and will automatically log into synapse when calling the login function.

batch/sc-batch-fargate.yaml
```diff
@@ -0,0 +1,200 @@
...
+            Value: !Ref SynapsePAT
+        NetworkConfiguration:
+          AssignPublicIp: ENABLED
+        ResourceRequirements:
+          - Type: VCPU
+            Value: 0.5
+          - Type: MEMORY
+            Value: 1024
```
thomasyu888, line 140: I think its fair to assume for most jobs that it won't take much CPU and memory, but maybe this should be configurable as well.

batch/sc-batch-fargate.yaml
```diff
@@ -0,0 +1,200 @@
...
+      RetryStrategy:
+        Attempts: 1
+      ContainerProperties:
+        Command: !Ref Command
+        Image: !Ref Image
+        Environment:
+          - Name: SynapsePAT
+            Value: !Ref SynapsePAT
+        NetworkConfiguration:
+          AssignPublicIp: ENABLED
```
thomasyu888, line 135: Why would cronjobs need public ip?

batch/sc-batch-fargate.yaml
```diff
@@ -0,0 +1,200 @@
...
+              Service:
+                - 'ecs-tasks.amazonaws.com'
+            Action:
+              - 'sts:AssumeRole'
+  LogGroup:
+    Type: AWS::Logs::LogGroup
+    Properties:
+      LogGroupName: !Join [ '', [ '/', { Ref: 'AWS::StackName' }, '/log' ] ]
+      RetentionInDays: 14
+  BatchTrigger:
```
thomasyu888, line 178: oh interesting... Maybe using fargate you need to use a lambda to trigger the job?  I didn't need to do that for AWS batch without fargate.

batch/sc-batch-fargate.yaml
```diff
@@ -0,0 +1,200 @@
...
+        Parameters:
+          - EnableSchedule
+          - Schedule
+Parameters:
+  Image:
+    Type: String
+    Description: >
+      The container image to run, can be from dockerhub or from AWS ECR repository.
+    Default: "debian:latest"
+    ConstraintDescription: >
```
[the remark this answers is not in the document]
zaro0508: yes, currently repos must be public. Will update description.

batch/sc-batch-fargate.yaml
zaro0508: yes exactly, using Fargate is almost like using lambda.  AWS will manage the compute instances that the tasks executes on so we don't have to manage it ourselves.

batch/sc-batch-fargate.yaml
zaro0508: Yup.  I initially tried to trigger the batch using a `AWS::Events::Rule` like it's done in Genie however that thing never triggered the my batch job.   I think it's a bug in AWS which i plan to confirm with AWS support.  In any case, I like the lambda approach better because it allows us to setup an API for users to manually trigger the job themselves.

batch/sc-batch-fargate.yaml
zaro0508: Env Vars are passed to the container on line 131 but seems like we may need more.

batch/sc-batch-fargate.yaml
zaro0508: removed

batch/sc-batch-fargate.yaml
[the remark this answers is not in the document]
zaro0508: description has been updated to explicitly state repos must be public

batch/sc-batch-fargate.yaml
zaro0508: Done

batch/sc-batch-fargate.yaml
```diff
@@ -0,0 +1,198 @@
...
+      Timeout:
+        AttemptDurationSeconds: 600
+      RetryStrategy:
+        Attempts: 1
+      ContainerProperties:
+        Command: !Ref Command
+        Image: !Ref Image
+        Environment:
+          - Name: SYNAPSE_AUTH_TOKEN
+            Value: !Ref SynapseAuthToken
```
ahayden, line 133: After talking with @zaro0508 , we think that this is the only input method for running a batch job through Service Catalog, particularly if the job requires the user to input a *scoped* access token... By taking input it was to be a stack parameter (though it could be stored somewhere else after). Other products can take secrets input after they are created, for example. If this is the only method, then we also need to restrict access to any non-admin roles to read stack parameters.

batch/sc-batch-fargate.yaml
```diff
@@ -0,0 +1,198 @@
...
+        Version: 2012-10-17
+        Statement:
+          - Effect: 'Allow'
+            Principal:
+              Service:
+                - 'ecs-tasks.amazonaws.com'
+            Action:
+              - 'sts:AssumeRole'
+     ManagedPolicyArns:
+       - 'arn:aws:iam::aws:policy/service-role/AmazonECSTaskExecutionRolePolicy'
```
ahayden, line 158: This service role should not be denied read on stack parameters, right?

batch/sc-batch-fargate.yaml
```diff
@@ -0,0 +1,198 @@
...
+        Version: 2012-10-17
+        Statement:
+          - Effect: 'Allow'
+            Principal:
+              Service:
+                - 'ecs-tasks.amazonaws.com'
+            Action:
+              - 'sts:AssumeRole'
+  LogGroup:
+    Type: AWS::Logs::LogGroup
```
ahayden, line 172: Would the cloudformation parameters or docker env get logged here?

batch/sc-batch-fargate.yaml
brucehoff: @ahayden Did you and @zaro0508 discuss the option of having the user put the secret into SSM (secure) parameter store, e.g. using the AWS CLI?  (The batch job would then retrieve the secret from there at run time.) I'm curious to know why you feel that option is infeasible.

batch/sc-batch-fargate.yaml
zaro0508: We now allow the user to pass an arbitrary list of ENV VARS

batch/sc-batch-fargate.yaml
zaro0508: It cannot read secrets in the stack parameters

batch/sc-batch-fargate.yaml
zaro0508: yes, if the user wants to log it.  It entirely up to the user and what they run on their docker image

zaro0508 (commented):
The latest update of this product is going to store secrets in the AWS secrets manager

batch/sc-batch-fargate.yaml
```diff
@@ -0,0 +1,200 @@
...
+    ParameterGroups:
+      - Label:
+          default: Container Image
+        Parameters:
+          - Image
+          - Command
+      - Label:
+          default: Synapse
+        Parameters:
+          - SynapsePAT
```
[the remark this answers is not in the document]
zaro0508: The latest update of this product is going to store secrets in the AWS secrets manager

batch/sc-batch-fargate.yaml
ahayden: `ServiceCatalogEndUsers` do not have rights to read LogGroups, so I'm not sure if this is as useful as you are thinking... Will they be asking admins to provider logs if they need debugging?

batch/sc-batch-fargate.yaml
```diff
@@ -0,0 +1,246 @@
...
+      ContainerProperties:
+        Command: !Ref Command
+        Image: !Ref Image
+        Environment: |
+          #!PyPlate
+          output = []
+          for tag in params['EnvVars']:
+             key, value = tag.split('=')
+             output.append({"Name": key, "Value": value})
+        Secrets:
```
ahayden, line 177: I like this design a lot... And I think it would be most straightforward to put all env vars in secretsmanager and to remove the `Environment` parameter to `ContainerProperties`. This would remove the dependency on PyPlate parsing and would require end users to write container jobs that interpret environment inputs in the way we want them to.

batch/sc-batch-fargate.yaml
```diff
@@ -0,0 +1,246 @@
...
+    Type: String
+    Default: cron(0 09 ? * MON *)  # Run every Monday at 9am
+    ConstraintDescription: "Use schedule format: https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html"
+  Secrets:
+    Type: String
+    NoEcho: true
+    Default: ""
+    Description: >
+      (Optional) The secrets passed to the docker container.
+    ConstraintDescription: 'Must be in "Key"="Value" form (i.e. "SECRET1":"Shh1","SECRET2":"Shh2")'
```
brucehoff, line 61: Typo:  Should be `"Key":"Value"`

brucehoff (changes requested):
Just one typo.

batch/sc-batch-fargate.yaml
zaro0508: Setup a mapping to allow users to select memory options that will map to CPU shares.  This will allow users to configure the memory for the container but will also constrain CPU shares to a matching pair of memory and VCPU shares.

batch/sc-batch-fargate.yaml
zaro0508: fixed

batch/sc-batch-fargate.yaml
zaro0508: I'm not a fan of this suggestion for the following reasons:
1.  There is a cost when adding vars to secrets manager which means it would be more costly
2. Putting vars that are not secrets into the secrets manager would make it more difficult to debug issues.
3. Using it that way is not how batch was designed.  It just feels like we would be swimming upstream to how batch should be used.

brucehoff (approved):

batch/sc-batch-fargate.yaml
```diff
@@ -0,0 +1,271 @@
...
+      - "1024"
+      - "2048"
+      - "4096"
+      - "10240"
+      - "18432"
+  EnableSchedule:
+    Description: >
+      true to run on a schedule, false to disable. If enabled a valid Schedule must be provided
+    Type: String
+    Default: false
```
thomasyu888, line 51: Oh interesting, this just run single process jobs too?  What does that look like as a SC product?

thomasyu888 (approved):

PR #266, opened 2021-12-15
[SC-306] Workaround for empty parameters

zaro0508:
The only workaround that I could find for allowing empty Secret and EnvVar parameters
is to not allow it at all.  Instead we set default values and user either needs to
accept the default values or replace them with their own VARS.  Setting the parameters
to an empty string will cause the deployment to fail.

### DIFF
--- a/batch/sc-batch-fargate.yaml
+++ b/batch/sc-batch-fargate.yaml
@@ -59,9 +59,9 @@ Parameters:
   Command:
     Type: String
     Description: >
-      A pipe seperated list of commands to pass to the container image (i.e. echo|hello|world)
+      A pipe separated list which is the command to pass to the container (i.e. echo|hello|world)
     ConstraintDescription: >
-      Must be a pipe delimited string of commands
+      Must be a pipe delimited list of strings
   Schedule:
     Description: >
       Schedule to execute the docker, can be a rate or a cron schedule. Format at
@@ -75,13 +75,13 @@ Parameters:
     Default: '"SECRET1":"Shh1"'
     Description: >
       The secrets passed to the docker container (i.e. "SECRET1":"Shh1","SECRET2":"Shh2")
-    ConstraintDescription: 'Must be in "Key":"Value" form'
+    ConstraintDescription: 'Must be in "Key":"Value" form, and cannot be omitted'
   EnvVars:
     Type: CommaDelimitedList
     Default: 'SCHEDULED_JOB_VAR1=one'
     Description: >
       The environment variables passed to the docker container (i.e. VAR1=One,VAR2=Two)
-    ConstraintDescription: 'Must be in Key=Value form'
+    ConstraintDescription: 'Must be in Key=Value form, and cannot be omitted'
 Transform: [PyPlate]
 Mappings:
   VpcuMemoryMap:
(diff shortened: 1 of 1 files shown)


batch/sc-batch-fargate.yaml
```diff
@@ -57,30 +57,30 @@ Parameters:
       - true
       - false
   Command:
-    Type: List<String>
+    Type: String
     Description: >
-      The list of commands to pass to the container image (i.e. ["echo", "hello", "world"])
+      A pipe seperated list of commands to pass to the container image (i.e. echo|hello|world)
```
brucehoff, line 62: typo: sep`a`rated
... and it's not a list of commands.  In the example 'world' isn't a command.  Instead you could say something like, 'A pipe separated list which is the command to pass to the container.'

batch/sc-batch-fargate.yaml
```diff
@@ -57,30 +57,30 @@ Parameters:
...
     ConstraintDescription: >
-      Must be a list of commands
+      Must be a pipe delimited string of commands
```
brucehoff, line 64: As per my above (nick-picky) comment, 'Must be a pipe delimited string.'

batch/sc-batch-fargate.yaml
```diff
@@ -57,30 +57,30 @@ Parameters:
...
     ConstraintDescription: "Use schedule format: https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html"
   Secrets:
     Type: String
     NoEcho: true
-    Default: ""
+    Default: '"SECRET1":"Shh1"'
     Description: >
-      (Optional) The secrets passed to the docker container (i.e. "SECRET1":"Shh1","SECRET2":"Shh2")
+      The secrets passed to the docker container (i.e. "SECRET1":"Shh1","SECRET2":"Shh2")
     ConstraintDescription: 'Must be in "Key":"Value" form'
```
brucehoff, line 78: ... ", and cannot be omitted."

batch/sc-batch-fargate.yaml
```diff
@@ -57,30 +57,30 @@ Parameters:
...
   EnvVars:
     Type: CommaDelimitedList
-    Default: ""
+    Default: 'SCHEDULED_JOB_VAR1=one'
     Description: >
-      (Optional) The environment variables passed to the docker container (i.e. VAR1=One,VAR2=Two)
+      The environment variables passed to the docker container (i.e. VAR1=One,VAR2=Two)
     ConstraintDescription: 'Must be in Key=Value form'
```
brucehoff, line 84: ", and cannot be omitted."

brucehoff (changes requested):

batch/sc-batch-fargate.yaml
zaro0508: done

batch/sc-batch-fargate.yaml
zaro0508: done

batch/sc-batch-fargate.yaml
zaro0508: done

batch/sc-batch-fargate.yaml
zaro0508: done

brucehoff (approved):
Looks great!

thomasyu888 (approved):

PR #268, opened 2022-01-04
quotes for AllowedPattern should only be used in json

zaro0508:


zaro0508:
replacing this PR with PR #269

PR #270, opened 2022-01-10
[SC-373] Make Command parameter optional

zaro0508:
* A dockerfile may run the require commands so user may not
  need to pass in a command to run.  This makes the command an
  optional input parameter.
* Change the separater to space instead of pipe to match how the user
  passes in the command using the docker CLI,
  https://docs.docker.com/engine/reference/builder/#cmd

### DIFF
--- a/batch/sc-batch-fargate.yaml
+++ b/batch/sc-batch-fargate.yaml
@@ -60,9 +60,7 @@ Parameters:
   Command:
     Type: String
     Description: >
-      (Optional) A comma separated list which is the command to pass to the container (i.e. echo,hello,world).
-    ConstraintDescription: >
-      Must be either empty or a comma delimited list of strings.
+      (Optional) The command that's passed to the docker container (i.e. echo hello world).
     Default: ""
   Schedule:
     Description: >
@@ -201,7 +199,7 @@ Resources:
       RetryStrategy:
         Attempts: 1
       ContainerProperties:
-        Command: !If [HasCommand, !Split [ "," , !Ref Command ], !Ref 'AWS::NoValue']
+        Command: !If [HasCommand, !Ref Command, !Ref 'AWS::NoValue']
         Image: !Ref Image
         Environment: |
           #!PyPlate
--- a/batch/sc-batch-fargate.yaml
+++ b/batch/sc-batch-fargate.yaml
@@ -60,7 +60,7 @@ Parameters:
   Command:
     Type: String
     Description: >
-      (Optional) The command that's passed to the docker container (i.e. echo hello world).
+      (Optional) The command that's passed to the docker container (e.g. echo hello world).
     Default: ""
   Schedule:
     Description: >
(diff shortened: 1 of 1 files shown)


batch/sc-batch-fargate.yaml
```diff
@@ -60,11 +60,10 @@ Parameters:
   Command:
     Type: String
     Description: >
-      A pipe separated list which is the command to pass to the container (i.e. echo|hello|world)
+      (Optional) A comma separated list which is the command to pass to the container (i.e. echo,hello,world).
```
brucehoff, line 63: I believe that Docker creates the run command as a space separated string of the (pipe or comma separated) tokens you give it, e.g. "echo,hello,world" -> "docker run <image> echo hello world".  So if you passed "echo hello world,foo,bar" Docker would run "docker run <image> echo hello world foo bar".  So why now just let an entire command be the first and only element of the Command array?  Let the user enter "echo hello world" and simply pass that through to Fargate/Docker.  It would make the Command param' much more readable in the UI.

brucehoff (changes requested):

batch/sc-batch-fargate.yaml
zaro0508: I’m just not sure it’s the most desirable option.  Docker natively prefers a comma separated list for command so why should we go with space separated?  it’s documented right here: https://docs.docker.com/engine/reference/builder/#cmd
```
The CMD instruction has three forms:

CMD ["executable","param1","param2"] (exec form, this is the preferred form)
CMD ["param1","param2"] (as default parameters to ENTRYPOINT)
CMD command param1 param2 (shell form)
```
i’m not opposed to spaced separated i’m just wondering why docker prefers comma?  I'm guessing that there's a good reason for it therefore i prefer to pass in with the format that docker prefers.

ahayden: I vote shell form for humans

thomasyu888: I think theres a different when executing the command in the Dockerfile vs when doing `docker run`.  For instance, you wouldn't do:

```
docker run sagebionetworks/synapsepythonclient synapse,get,-r,syn1234
```

You would do
```
docker run sagebionetworks/synapsepythonclient synapse get -r syn1234
```

I vote shell form

zaro0508: changed to shell form

batch/sc-batch-fargate.yaml
```diff
@@ -60,11 +60,8 @@ Parameters:
   Command:
     Type: String
     Description: >
-      A pipe separated list which is the command to pass to the container (i.e. echo|hello|world)
-    ConstraintDescription: >
-      Must be a pipe delimited list of strings
-    Default: "echo|hello|world"
-    AllowedPattern: "^(?!s*$).+"
+      (Optional) The command that's passed to the docker container (i.e. echo hello world).
```
brucehoff, line 63: Grammar:  Should be "e.g., echo hello world", not "i.e. echo hello world".

brucehoff (changes requested):

batch/sc-batch-fargate.yaml
zaro0508: fixed

PR #274, opened 2022-03-21
Remove integration with VPN

zaro0508:
We no longer use the VPN to access instances in the service catalog.
Our instructions[1] provide directions to access the instance using
the AWS session manager which is a more secure form of access.

[1] https://help.sc.sageit.org/sc/Service-Catalog-Provisioning.938836322.html#ServiceCatalogProvisioning-SSMaccesstoanInstance


brucehoff (approved):

zaro0508:
Hold off on merging, we may want PR #275 instead

PR #275, opened 2022-03-22
[IT-1234] AWS Security group for TGW

zaro0508:
We use the transit gateway to route traffic between the hub
and spoke VPCs. To allow access to resources in a spoke VPC we need
to give the hub VPC access to the resources. This adds an SG ingress
to resources provisioned in our spoke VPCs which will allow VPN users
access to the resources in the spoke VPCs.


zaro0508:
This PR is the converse of PR #274

xschildw (approved):

ahayden (approved):

PR #282, opened 2022-10-12
[IT-2132] Propagate tags from instance to volumes

jesusaurus:
Propagate the Service Catalog tags from the EC2 instance to its associated EBS volumes.

### DIFF
--- a/ec2/sc-ec2-linux-docker.yaml
+++ b/ec2/sc-ec2-linux-docker.yaml
@@ -269,6 +269,7 @@ Resources:
           #!/bin/bash
           /opt/aws/bin/cfn-init -v --stack ${AWS::StackName} --resource EC2Instance --configsets SetupCfn,SetEnv --region ${AWS::Region}
           /opt/aws/bin/cfn-signal -e $? --stack ${AWS::StackName} --resource EC2Instance --region ${AWS::Region}
+      PropagateTagsToVolumeOnCreation: true
       Tags:
         - Key: Name
           Value: !Ref 'AWS::StackName'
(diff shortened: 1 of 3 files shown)


xschildw (approved):

zaro0508 (changes requested):
The `PropagateTagsToVolumeOnCreation` feature was introduced after we created these files products.  We had to create a custom cloudformation resource that would propagate tags to volumes and I still see that it's in these files.  Take a look at `custom:SynapseTagger` below.  That references cfn-cr-synapse-tagger which will run [set_instance_tags.py](https://github.com/Sage-Bionetworks-IT/cfn-cr-synapse-tagger/blob/master/set_tags/set_instance_tags.py#L89). 

I think this is the right refactor,  use `PropagateTagsToVolumeOnCreation` going forward however I think we should also remove that functionality from `cfn-cr-synapse-tagger`.  I'm thinking we should remove it from the custom resource before adding this so there's no conflict going on.

zaro0508:
@jesusaurus have we verified whether `PropagateTagsToVolumeOnCreation` AND the apply tags in `cfn-cr-synapse-tagger` are both additive and are not going to replace existing tags?

jesusaurus:
@zaro0508 I was able to confirm that the tags in `cfn-cr-synapse-tagger` are strictly additive, but I haven't been able to think of a way to force a race condition during resource creation to test `PropagateTagsToVolumeOnCreation` and the docs don't specify.

zaro0508:
i think it's safe from race condition because the TagInstance `DependsOn: "InstanceProfile"` makes cloudformation execute `EC2Instance` and  `TagInstance` sequentially

zaro0508 (approved):

zaro0508:
@jesusaurus you'll need to create a Tag (release) and an additional update to the [service catalog EC2 product yaml files](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/sceptre/scipool/config/develop/sc-product-ec2-linux-docker.yaml#L17-L41) to use the new version of the products.

PR #287, opened 2022-12-11
[SC-404] Fix exposure

zaro0508:
Use ENV VAR to store variables so it won't get captured in cloudformation log files.


xschildw (approved):
So $env:JC_SERVICE_API_KEY' does not get expanded on the cmdline?

zaro0508:
This will not work, review the jira issue for more details.

PR #288, opened 2022-12-13
[SC-404] Stop JC exposure

zaro0508:
Remove automated association of windows instance to jumpcloud user. This still allows us to do the JC user and instance association manually in jumpcloud.


xschildw (approved):
Good point, we should not be talking dozens of instances...

jesusaurus (approved):

brucehoff (approved):

PR #301, opened 2023-07-31
SC-383: Update utility version

xschildw:
This PR updates the version of service-catalog-utility (to pick up fix on apache .conf rewrite).

Depends on https://github.com/Sage-Bionetworks/service-catalog-utils/pull/22 .

### DIFF
--- a/templates/ec2/sc-ec2-linux-jumpcloud-notebook.yaml
+++ b/templates/ec2/sc-ec2-linux-jumpcloud-notebook.yaml
@@ -24,7 +24,7 @@ Metadata:
 Mappings:
   NotebookTypes:
     Rstudio:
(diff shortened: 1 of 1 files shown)


brucehoff (approved):

zaro0508 (changes requested):
i think we should just remove this apache rewrite step altogether since it's already being done in [packer-rstudio image](https://github.com/Sage-Bionetworks-IT/packer-rstudio/blob/master/src/proxy.conf).  i think setting it up in packer-rstudio is easier and more maintainable than setting it up with a custom script here.

xschildw:
Can't do, the file in the image has a placeholder that's replaced at launch time by the instance id (at least for apache conf).

brucehoff (approved):

PR #318, opened 2023-12-15
IT-3317

brucehoff:
Add template for dockerized service catalog notebook product.

This is modeled after the current RStudio notebook template, however instead of using the RStudio AMI it simply uses the base Ubuntu AMI and then runs the reverse proxy and notebook software as containers.  This modularity allows the product to have the option of running either an RStudio or Jupyter notebook.

Depends on https://github.com/Sage-Bionetworks-IT/notebook-reverse-proxy/pull/1


ec2/sc-ec2-linux-docker-notebook.yaml
```diff
@@ -0,0 +1,513 @@
...
+      Parameters:
+        # include is @ https://github.com/Sage-Bionetworks/admincentral-infra/blob/master/templates/cfn-service-catalog-snippets.yaml
+        Location: s3://cfn-snippets-bucket-cloudformationsnippetsbucket-elu83sv8ocdz/scipool-sc-lib-infra/ScAccountToExportMappping.yaml
+  NotebookTypeToInitParams:
+    Jupyter:
+      NoteBookTypeForDocker: jupyter
+      ConfigSet: StartContainersJupyter
+      DockerNotebookContainerName: jupyter-notebook
+      DockerNetworkName: proxy-net
+      DockerImage: quay.io/jupyter/base-notebook:python-3.11
```
zaro0508, line 37: hmm, would it make sense to allow users to pick a DockerImage instead of a NotebookType?  We can assume the type of notebook the user wants from the docker image that's select, correct?  parameterizing around DockerImage would allow us provide a more specific option.

ec2/sc-ec2-linux-docker-notebook.yaml
```diff
@@ -0,0 +1,513 @@
...
+            start_docker_network:
+              command: "docker network inspect ${NETWORK_NAME} ||  \
+                docker network create --driver bridge ${NETWORK_NAME}"
+              env:
+                NETWORK_NAME:
+                  'Fn::FindInMap': [NotebookTypeToInitParams, !Ref NotebookType, DockerNetworkName]
+        start_reverse_proxy:
+          commands:
+            start_reverse_proxy:
+              command: "docker start reverse-proxy || \
```
zaro0508, line 330: `docker start..` or `docker run..`?  i'm not sure why you need `docker start`?  doesn't `docker run..` suffice?  also why run one command with `sudo` and the other without?

ec2/sc-ec2-linux-docker-notebook.yaml
```diff
@@ -0,0 +1,513 @@
...
+                ${DOCKER_IMAGE}"
+              env:
+                NETWORK_NAME:
+                  'Fn::FindInMap': [NotebookTypeToInitParams, !Ref NotebookType, DockerNetworkName]
+                NOTEBOOK_CONTAINER_NAME:
+                  'Fn::FindInMap': [NotebookTypeToInitParams, !Ref NotebookType, DockerNotebookContainerName]
+                DOCKER_IMAGE:
+                  'Fn::FindInMap': [NotebookTypeToInitParams, !Ref NotebookType, DockerImage]
+                EC2_WORK_DIR:
+                  'Fn::FindInMap': [NotebookTypeToInitParams, !Ref NotebookType, EC2WorkDir]
```
zaro0508, line 399: there's a `JUPYTER_WORK_DIR` env var `start_jupyter_notebook` but not one here?  for consistency maybe create a generic one called `APP_WORK_DIR` for both?

ec2/sc-ec2-linux-docker-notebook.yaml
```diff
@@ -0,0 +1,513 @@
...
+
+    Properties:
```
zaro0508, line 401: missing `PropagateTagsToVolumeOnCreation: true`

ec2/sc-ec2-linux-docker-notebook.yaml
```diff
@@ -0,0 +1,513 @@
...
+
+  LinuxInstance:
+    Type: AWS::EC2::Instance
+    Metadata:
+      cfn-lint:
+        config:
+          ignore_checks:
+            - E1022
+            - W1011
+      'AWS::CloudFormation::Init':
```
zaro0508, line 307: In the current SC deployment of rstudio we also [setup synapse credentials for apache and rstudio](https://github.com/Sage-Bionetworks/service-catalog-library/blob/master/templates/ec2/sc-ec2-linux-jumpcloud-notebook.yaml#L299-L333) and I believe those things depend on [set_env_vars](https://github.com/Sage-Bionetworks/service-catalog-library/blob/master/templates/ec2/sc-ec2-linux-jumpcloud-notebook.yaml#L285-L299).  I think we still need to do that somewhere in here?

zaro0508 (changes requested):

ec2/sc-ec2-linux-docker-notebook.yaml
```diff
@@ -0,0 +1,513 @@
...
+    'Fn::Transform':
+      Name: 'AWS::Include'
+      Parameters:
+        # include is @ https://github.com/Sage-Bionetworks/admincentral-infra/blob/master/templates/cfn-service-catalog-snippets.yaml
+        Location: s3://cfn-snippets-bucket-cloudformationsnippetsbucket-elu83sv8ocdz/scipool-sc-lib-infra/ScAccountToExportMappping.yaml
+  NotebookTypeToInitParams:
+    Jupyter:
+      NoteBookTypeForDocker: jupyter
+      ConfigSet: StartContainersJupyter
+      DockerNotebookContainerName: jupyter-notebook
```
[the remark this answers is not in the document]
jesusaurus: If we make this more generic, I think we can remove it from the mapping entirely.

ec2/sc-ec2-linux-docker-notebook.yaml
```diff
@@ -0,0 +1,513 @@
...
+                -e NOTEBOOK_HOST=${NOTEBOOK_CONTAINER_NAME} \
+                -e AWS_REGION=${AWS_REGION} \
+                -e SERVICE_CATALOG_PREFIX=${SERVICE_CATALOG_PREFIX} \
+                -e SSM_PARAMETER_SUFFIX=${SSM_PARAMETER_SUFFIX} \
+                ghcr.io/Sage-Bionetworks-IT/notebook-reverse-proxy-${NOTEBOOK_TYPE}:1.0"
+              env:
+                AWS_REGION: !Ref AWS::Region
+                NETWORK_NAME:
+                  'Fn::FindInMap': [NotebookTypeToInitParams, !Ref NotebookType, DockerNetworkName]
+                NOTEBOOK_CONTAINER_NAME:
```
jesusaurus, line 345: If this is something generic like `notebook`, it won't need `FindInMap`

ec2/sc-ec2-linux-docker-notebook.yaml
```diff
@@ -0,0 +1,513 @@
...
+      ConfigSet: StartContainersRstudio
+      DockerNotebookContainerName: rstudio-notebook
+      DockerNetworkName: proxy-net
+      DockerImage: rocker/rstudio:4.3.2
+      EC2WorkDir: /home/ssm-user/workdir
+
+Parameters:
+
+  EC2InstanceType:
+    AllowedValues:
```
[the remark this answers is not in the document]
brucehoff: Yes, good catch.  To fix this, I "diffed" the old and new versions of the `sc-ec2-linux-jumpcloud-notebook.yaml` I copied from, and added all the changes into this file, including the list of allowed instance types.

ec2/sc-ec2-linux-docker-notebook.yaml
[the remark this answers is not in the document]
brucehoff: Either way works.  If you all feel strongly about minimizing the mapping parameters, we can remove both `DockerNotebookContainerName` and `DockerNetworkName`, instead using generic values.

ec2/sc-ec2-linux-docker-notebook.yaml
brucehoff: > hmm, would it make sense to allow users to pick a DockerImage instead of a NotebookType?
Yes, but let's refer to each image we allow as a "NotebookType" in this template. E.g., we could add a new type, `RStudioForBMGFki`, which would be a distinct type based on RStudio.

> We can assume the type of notebook the user wants from the docker image that's select, correct?
I would not want to try to guess whether an image is RStudio or Jupyter from the image name. 

> parameterizing around DockerImage would allow us provide a more specific option.
Yes, each "NotebookType" has a `DockerImage` parameter which provides a specific notebook option.

ec2/sc-ec2-linux-docker-notebook.yaml
```diff
@@ -0,0 +1,513 @@
...
+          ignore_checks:
+            - E1022
+            - W1011
+      'AWS::CloudFormation::Init':
+        configSets:
+          StartContainersJupyter:
+            - start_docker_network
+            - start_reverse_proxy
+            - make_work_dir
+            - start_jupyter_notebook
```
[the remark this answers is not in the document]
brucehoff: > both will run?
No:  `start_jupyter_notebook` is in the `StartContainersJupyter` config' set, while `start_rstudio_notebook` is in the `StartContainersRstudio` config' set.

> don't we want to run only one of these depending on what the user sets for NotebookType?
Yes: The `StartContainersJupyter` config' set will run for the Jupyter notebook type and the `StartContainersRstudio` config' set will run for the RStudio notebook type.

ec2/sc-ec2-linux-docker-notebook.yaml
brucehoff: > docker start.. or docker run..? i'm not sure why you need docker start? doesn't docker run.. suffice?
It's for idempotency:  If the container already exists then this command will restart it but if it doesn't then the 'docker run' command will pull down the image, create the container, and start it up.

> also why run one command with sudo and the other without?
Neither should have 'sudo'.  I will update accordingly.

ec2/sc-ec2-linux-docker-notebook.yaml
```diff
@@ -0,0 +1,513 @@
...
+                  'Fn::FindInMap': [NotebookTypeToInitParams, !Ref NotebookType, DockerNetworkName]
+        start_reverse_proxy:
+          commands:
+            start_reverse_proxy:
+              command: "docker start reverse-proxy || \
+                sudo docker run -d -p 443:443 \
+                --name reverse-proxy \
+                --network ${NETWORK_NAME} \
+                --restart unless-stopped \
+                -e EC2_INSTANCE_ID=$(curl http://169.254.169.254/latest/meta-data/instance-id) \
```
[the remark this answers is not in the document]
brucehoff: Apparently we do not have to update.  I have tried this out in our dev Service Catalog and it works fine.

ec2/sc-ec2-linux-docker-notebook.yaml
```diff
@@ -0,0 +1,513 @@
...
+  ConnectionInstructions:
+    Description: 'Guidelines on connecting to instances'
+    Value: "https://sagebionetworks.jira.com/wiki/spaces/IT/pages/996376579/Connect+to+Provisioned+Instances"
+  ConnectionURI:
+    Description: 'Starts a shell session in the AWS Console'
+    Value: !Sub "https://${AWS::Region}.console.aws.amazon.com/systems-manager/session-manager/${LinuxInstance}?region=${AWS::Region}"
+  EC2ConsoleURI:
+    Description: 'Check your instance status with this link to the AWS Console'
+    Value: !Sub "https://console.aws.amazon.com/ec2/v2/home?region=${AWS::Region}#Instances:search=${LinuxInstance}"
+  NotebookConnectionURI:
```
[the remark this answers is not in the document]
brucehoff: Yes.

ec2/sc-ec2-linux-docker-notebook.yaml
```diff
@@ -0,0 +1,513 @@
...
+    Properties:
+      Path: /
+      Roles:
+        - !Ref 'InstanceRole'
+
+  # The EC2 instance writes to it, but this is introduced for the cleanup of the parameter
+  InstanceSSMParameter:
+    Type: AWS::SSM::Parameter
+    Properties:
+      # This will need to be synced with the `packer-rstudio` project
```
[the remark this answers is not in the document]
brucehoff: No, this is just a stray comment, copied from the current rstudio notebook template.  The relevant logic from the packer-rstudio image is incorporated in the reverse-proxy container.  I have removed the comment and also defined variables for `/service-catalog/synapse/cred` and `oidc-accesstoken`.

ec2/sc-ec2-linux-docker-notebook.yaml
```diff
@@ -0,0 +1,513 @@
...
+            - "kms:Get*"
+            - "kms:Delete*"
+            - "kms:ScheduleKeyDeletion"
+            - "kms:CancelKeyDeletion"
+          Resource: "*"
+
+  SSMParameterEncryptionKeyAlias:
+    Type: AWS::KMS::Alias
+    Properties:
+      # This will need to be synced with the `packer-rstudio` project
```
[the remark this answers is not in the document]
brucehoff: Please see my response, above.

ec2/sc-ec2-linux-docker-notebook.yaml
brucehoff: Yes, you are right.  The environment variables, `SYNAPSE_TOKEN_AWS_SSM_PARAMETER_NAME` and `AWS_DEFAULT_REGION` have to be set, to allow the Synapse client to retrieve the Synapse access token from parameter store. (See [this](https://raw.githubusercontent.com/Sage-Bionetworks/service-catalog-utils/v2.0.3/linux/opt/sage/bin/rstudio_synapse_cred_env.sh) script.).  I have updated this PR to add the two env var's to both the RStudio and Jupyter notebooks.

ec2/sc-ec2-linux-docker-notebook.yaml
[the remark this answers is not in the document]
zaro0508: it's because you are building on top of an older AWS AMI,  the [IMDSV1 will be removed in mid 2024](https://aws.amazon.com/blogs/aws/amazon-ec2-instance-metadata-service-imdsv2-by-default/).  We should either do this now or in a follow on PR.

ec2/sc-ec2-linux-docker-notebook.yaml
zaro0508: > I would not want to try to guess whether an image is RStudio or Jupyter from the image name
why would we need to guess?  either `jupyter` or `rstudio` is in the path of the `DockerImage` parameter, `quay.io/jupyter/base-notebook:python-3.11` and `rocker/rstudio:4.3.2`

brucehoff: In the future we may add other RStudio and Jupyter images and the author may not adhere to the naming convention, e.g., they might want to make a Docker repo' called `sagebionetworks/my-project-notebook`.

templates/ec2/sc-ec2-linux-docker-notebook.yaml
```diff
@@ -0,0 +1,491 @@
...
+                mkdir -p ${EC2_WORK_DIR}
+                sudo chmod 777 ${EC2_WORK_DIR}
+              env:
+                EC2_WORK_DIR:
+                  'Fn::FindInMap': [NotebookTypeToInitParams, !Ref NotebookType, EC2WorkDir]
+        start_jupyter_notebook:
+          commands:
+            start_notebook:
+              command: |
+                EC2_INSTANCE_ID=$(curl http://169.254.169.254/latest/meta-data/instance-id)
```
zaro0508, line 320: shouldn't this env var be passed into the docker container (with `-e` flag) as well?

templates/ec2/sc-ec2-linux-docker-notebook.yaml
```diff
@@ -0,0 +1,491 @@
...
+                  'Fn::FindInMap': [GlobalVars, SSMParameterPrefix, Value]
+                SSM_PARAMETER_SUFFIX:
+                  'Fn::FindInMap': [GlobalVars, SSMParameterSuffix, Value]
+                AWS_REGION:
+                  'Fn::Sub': ${AWS::Region}
+        start_rstudio_notebook:
+          commands:
+            start_notebook:
+              command: |
+                EC2_INSTANCE_ID=$(curl http://169.254.169.254/latest/meta-data/instance-id)
```
zaro0508, line 353: same here?

templates/ec2/sc-ec2-linux-docker-notebook.yaml
```diff
@@ -0,0 +1,491 @@
...
+        'Fn::Sub': '${AWS::Region}-cfn-cr-alb-rule-FunctionArn'
+      InstanceId: !Ref LinuxInstance
+      TargetGroupArn: !Ref EC2TargetGroup
+      ListenerArn: !ImportValue
+        'Fn::Sub': '${AWS::Region}-alb-notebook-access-ALBListenerARN'
+  TagInstance:
+    DependsOn: "InstanceProfile"
+    Type: Custom::SynapseTagger
+    Properties:
+      ServiceToken: !ImportValue
```
zaro0508, line 463: This lambda will take all of the [ENV VARs defined in make_env_vars_file.sh file](https://github.com/Sage-Bionetworks/service-catalog-library/blob/master/templates/ec2/sc-ec2-linux-jumpcloud-notebook.yaml#L285-L298) and apply them as EC2 instance tags.  Some of those tags are very important to how the service catalog works.  For example the value for the `ACCESS_APPROVED_ROLEID` tag is what is used to restrict instance access to the user who provisioned it.  Since the make_env_vars_file.sh is no longer setup in this file i'm going to assume that those tags are no longer getting set.  I think we'll need `make_env_vars_file.sh` or find an alternative way to set those tags.

zaro0508 (changes requested):

templates/ec2/sc-ec2-linux-docker-notebook.yaml
brucehoff: I don't think so: From what I can tell the values to be passed into the container are `SYNAPSE_TOKEN_AWS_SSM_PARAMETER_NAME` and `AWS_DEFAULT_REGION`. (See [this](https://raw.githubusercontent.com/Sage-Bionetworks/service-catalog-utils/v2.0.3/linux/opt/sage/bin/rstudio_synapse_cred_env.sh) script.) The EC2 instance ID is embedded in `SYNAPSE_TOKEN_AWS_SSM_PARAMETER_NAME` but not otherwise used in the Notebook.  If you feel I am mistaken please let me know.

templates/ec2/sc-ec2-linux-docker-notebook.yaml
brucehoff: Please see my reply above.

templates/ec2/sc-ec2-linux-docker-notebook.yaml
brucehoff: I see. Will add to the config.

zaro0508 (approved):

PR #323, opened 2024-01-25
IT-3317: Correct volume mounts; update RStudio container to v1.0.1

brucehoff:
The previous version had two mistakes in mounting a folder from the parent EC2 into the notebook container:
- The working directory for Jupyter is `/home/jovyan/work` not `/home/jovyan/workdir`.
- In RStudio we mounted the (empty) EC2 folder to `/home/rstudio` which overwrote all the packages, venv, etc. that were intalled in the user homedir. This update simply mounts the host folder into the (otherwise non-existent) `/home/rstudio/work`.

Just for consistency we rename the host folder from `/home/ssm-user/workdir` to `/home/ssm-user/work`.

Depends on https://github.com/Sage-Bionetworks-IT/rstudio-service-catalog/pull/2, which will also have to be tagged with `v1.0.1` for this PR to work.

### DIFF
--- a/templates/ec2/sc-ec2-linux-docker-notebook.yaml
+++ b/templates/ec2/sc-ec2-linux-docker-notebook.yaml
@@ -34,11 +34,13 @@ Mappings:
       ConfigSet: StartContainersJupyter
       DockerImage: quay.io/jupyter/base-notebook:python-3.11
       EC2WorkDir: /home/ssm-user/work
+      NotebookWorkDir: /home/jovyan/work
     Rstudio:
       NoteBookTypeForDocker: rstudio
       ConfigSet: StartContainersRstudio
       DockerImage: ghcr.io/sage-bionetworks-it/rstudio-service-catalog:v1.0.1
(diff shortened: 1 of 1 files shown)


templates/ec2/sc-ec2-linux-docker-notebook.yaml
```diff
@@ -358,7 +358,7 @@ Resources:
                   'Fn::FindInMap': [NotebookTypeToInitParams, !Ref NotebookType, EC2WorkDir]
                 DOCKER_IMAGE:
                   'Fn::FindInMap': [NotebookTypeToInitParams, !Ref NotebookType, DockerImage]
-                JUPYTER_WORK_DIR: /home/jovyan/workdir
+                JUPYTER_WORK_DIR: /home/jovyan/work
```
zaro0508, line 361: might be better to give this a generic name, like `APP_WORK_DIR` and add it to the Jupyter and Rstudio mappings?

zaro0508 (approved):

PR #324, opened 2024-02-02
Windows uses x86 types

zaro0508:
The windows AMI doesn't seem to work with AMD instance types

This reverts commit 44b30a7988a104e8a834ad1d1abc27ef8001fabc for the windows products only.


zaro0508:
this is not the problem

PR #335, opened 2024-07-17
[IT-3523] Upload 'testing' branch for pre-merge testing

jesusaurus:
Upload changes to a 'testing' branch to S3 so that we can test them before merging to the master branch. The changes will be tested by referencing them in SC products in the dev account.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -20,7 +20,7 @@ jobs:
   deploy:
     name: Upload to Amazon S3
     runs-on: ubuntu-latest
-    if: ${{ github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v') }}
+    if: ${{ github.ref == 'refs/heads/master' || github.ref == 'refs/heads/testing' || startsWith(github.ref, 'refs/tags/v') }}
     needs: [ "validate" ]
     permissions:
       id-token: write


.github/workflows/deploy.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
   deploy:
     name: Upload to Amazon S3
     runs-on: ubuntu-latest
-    if: ${{ github.ref == 'refs/heads/master' || startsWith(github.ref, 'refs/tags/v') }}
+    if: ${{ github.ref == 'refs/heads/master' || github.ref == 'refs/heads/testing' || startsWith(github.ref, 'refs/tags/v') }}
```
zaro0508, line 23: we deploy from master on every change so i don't see how using `testing` branch would be different from using master branch?

also would deployment from testing branch be shown in https://ami.sageit.org/ ?  if so i'm not sure i like the idea of putting a bunch of testing AMIs on the same page as prod AMIs.  i think that might get kinda messy but i can be persuaded otherwise.

zaro0508 (changes requested):

.github/workflows/deploy.yaml
jesusaurus: the purpose of this PR is to enable pre-merge testing of packer changes, and I don't think iterating on those packer changes should require merging the intermediate change here.

I had no idea about ami.sageit.org, but I see there are already several test images listed (a few I built manually for testing, and a few from developing the docker notebook product). how do you feel about a change to the dashboard script to ignore AMIs with 'test' in the name?

zaro0508:
> the purpose of this PR is to enable pre-merge testing of packer changes, and I don't think iterating on those packer changes should require merging the intermediate change here.

yes, good point.  However I'm not sure why you even need to involve the CI deployment at all to accomplish pre-merge testing.  While in development, can you not just use your SSO account to deploy test AMIs to AWS and make them available for you to do pre-merge testing?


> how do you feel about a change to the dashboard script to ignore AMIs with 'test' in the name?

I like this idea.

.github/workflows/deploy.yaml
zaro0508: > the purpose of this PR is to enable pre-merge testing of packer changes, and I don't think iterating on those packer changes should require merging the intermediate change here.

yes, good point.  However I'm not sure why you even need to involve the CI deployment at all to accomplish pre-merge testing.  While in development, can you not just use your SSO account to deploy test AMIs to AWS and make them available for you to do pre-merge testing?


> how do you feel about a change to the dashboard script to ignore AMIs with 'test' in the name?

I like this idea.

jesusaurus: done in Sage-Bionetworks-IT/imagecentral-dashboard#11

ConsoleCatzirl: My main goal with this effort is to automate the steps of the testing process to reduce human error. I found myself writing several bash scripts to make sure I got the commands right and they deployed to the correct account. And if I'm going to automate those steps for myself, I figured I should generalize it for our development process. Automating it may not be strictly necessary, but I hope to improve my own ability to iterate on changes to service catalog products.

PR #342, opened 2025-04-24
IT-4369: Pre-install the policy for SSM patching

xschildw:
This PR adds the AWS-manged policy 'AWSQuickSetupPatchPolicyBaselineAccess' the the instance profiles
so that it's under control of Cloudformation and can be detached from the role when an instance is terminated.

NB: Hopefully adding is OK and will be a no-op since the policy has already been added manually.


ConsoleCatzirl (approved):

ConsoleCatzirl:
thanks for looking into this

